PR #6, opened 2025-11-19
Add chained FAS operations until fixed-point

mschlaipfer:
*Issue #, if available:*

*Description of changes:* Add FAS operations until fixed-point, not just a single one. Ensure we detect cycles to avoid infinite loops. Changes the datatype for the FAS operation context so we can implement `Hash`.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
@@ -86,7 +86,6 @@ impl ResourceMatcher {
     /// This method safely expands FAS operations by iteratively processing new operations
     /// until no more new operations are discovered (fixed point reached).
     /// It includes cycle detection to prevent infinite loops.
-    // TODO: If an operation occurs with and without context we may only keep the one with context.
     fn expand_fas_operations_to_fixed_point(
         &self,
         initial: FasOperation,


iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -89,13 +81,88 @@ impl ResourceMatcher {
             .cloned()
     }
 
-    fn make_condition(context: &HashMap<String, String>) -> Vec<Condition> {
+    /// Expand FAS operations to a fixed point, avoiding infinite loops from cycles
+    ///
+    /// This method safely expands FAS operations by iteratively processing new operations
+    /// until no more new operations are discovered (fixed point reached).
+    /// It includes cycle detection to prevent infinite loops.
+    // TODO: If an operation occurs with and without context we may only keep the one with context.
```
weibenz1: > If an operation occurs with and without context we may only keep the one with context.

It might cause access denial for the ones needed no context.

iam-policy-autopilot-policy-generation/src/enrichment/operation_fas_map.rs
```diff
@@ -44,30 +44,72 @@ pub(crate) struct OperationFasMap {
...
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq, Hash)]
 pub(crate) struct FasOperation {
     #[serde(rename = "Operation")]
     operation: String,
     #[serde(rename = "Service")]
     service: String,
-    #[serde(rename = "Context")]
-    pub(crate) context: HashMap<String, String>,
+    #[serde(rename = "Context", deserialize_with = "deserialize_context_map")]
+    pub(crate) context: Vec<Context>,
```
weibenz1, line 100: Q: why is vector better than Map in this case?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: Ah, good catch, will remove

iam-policy-autopilot-policy-generation/src/enrichment/operation_fas_map.rs
mschlaipfer: Need `Hash`, I found it cleaner to implement a `Context` type.

PR #15, opened 2025-11-21
Fix some low-hanging performance issues

mschlaipfer:
*Issue #, if available:*

*Description of changes:*
* Moved `ResourceDirectCallsExtractor::new(service_index)` out of the loop
* Instead of iterating over patterns, crated an ast-grep config with `any`

We should make sure this doesn't regress on our benchmark set before merging.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/extraction/python/boto3_resources_model.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/python/boto3_resources_model.rs
@@ -253,8 +253,14 @@ impl Boto3ResourcesRegistry {
             }
         };
 
+        // Build service versions map once for all services
+        let service_versions = EmbeddedBoto3Data::build_service_versions_map();
+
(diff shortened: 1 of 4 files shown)


iam-policy-autopilot-policy-generation/src/extraction/python/boto3_resources_model.rs
```diff
@@ -719,54 +728,6 @@ mod tests {
         assert_eq!("CreateBucket".to_case(Case::Snake), "create_bucket");
     }
 
-    #[test]
-    fn test_load_dynamodb_model_from_embedded() {
-        let result = Boto3ResourcesModel::load_from_embedded("dynamodb");
```
adpaco, line 724: We are removing these tests but not adding new ones for `load_from_embedded_with_versions`. Could we have at least one? Maybe using `dynamodb` and `s3` so you're recreating these tests.

PR #16, opened 2025-11-21
Add resource cutoff

mschlaipfer:
*Issue #, if available:*

*Description of changes:* Do not add resources when there are more than `RESOURCE_CUTOFF` (currently set to `5`). This prevents creating overly verbose policies at the expense of less precision. Particularly `ec2:CreateTags` was affected (see https://servicereference.us-east-1.amazonaws.com/v1/ec2/ec2.json)

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
@@ -27,6 +27,9 @@ pub(crate) struct ResourceMatcher {
     fas_maps: OperationFasMaps,
 }
 
+// TODO: Make this configurable: https://github.com/awslabs/iam-policy-autopilot/issues/19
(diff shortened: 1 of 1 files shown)


iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -222,6 +224,13 @@ impl ResourceMatcher {
                                         &action.name,
                                         &service_reference,
                                     )?;
+                                // TODO: Make this configurable
```
adpaco: I recommend opening an issue and linking here so we can track the work while it's pending.

mschlaipfer: Created https://github.com/awslabs/iam-policy-autopilot/issues/19 and linked in code.

PR #20, opened 2025-11-24
Add service hints feature

mschlaipfer:
*Issue #, if available:*

*Description of changes:*
* Added `--service-hints` feature with service synonym matching and "did you mean" suggestion (based on edit-distance <= 2)
* Refactored service model index building (no need to call `discover_services` anymore) and added caching to it
* Added service renaming for Smithy model names to Botocore model names used for building the service synonyms: In Smithy it is e.g. `cloudwatch-logs`, but in botocore it is `logs` (etc.). Furthermore, the import in Go is `cloudwatchlogs`, in JS/TS `cloudwatch-logs`. All of these are valid service-hints. I created the rename map by analyzing the `serviceId`/`sdkId` in the model files. We should eventually add a script which checks that this map is still up-to-date during CI.


iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -18,13 +18,22 @@ pub struct GeneratePolicyConfig {
     pub disable_file_system_cache: bool,
 }
 
+/// Service hints for filtering SDK method calls
+#[derive(Debug, Clone)]
+pub struct ServiceHints {
```
weibenz1, line 23: nit: This is more like ServiceHitTargets etc.

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -18,13 +18,22 @@ pub struct GeneratePolicyConfig {
...
+    /// List of AWS service names to filter by
+    pub service_names: Vec<String>,
```
weibenz1, line 25: would set make more sense?

PR #29, opened 2025-11-25
fix: correct handling of positional parameters for boto3 utils and r…

afilieri-aws:
*Issue #, if available:*

*Description of changes:*

Correct handling of positional parameters for boto3 utils and resource methods to reduce false positives.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/extraction/python/common/argument_extractor.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/python/common/argument_extractor.rs
@@ -19,6 +19,11 @@ impl ArgumentExtractor {
         let mut parameter_position = 0; // Track actual parameter position (excluding commas)
 
         for arg_node in args_nodes {
+            // Filter out comment nodes
+            if arg_node.kind() == "comment" {
+                continue;
+            }
+
             let arg_text = arg_node.text().to_string();
 
             // Filter out comma nodes - ast-grep captures commas as separate nodes
--- a/iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
@@ -101,27 +101,27 @@ impl Extractor for PythonExtractor {
         for extractor_result in extractor_results.iter_mut() {
             match extractor_result {
                 ExtractorResult::Python(ast, method_calls) => {
-                    // Extract resource direct calls (with ServiceModelIndex access)
+                    // STEP 1: Disambiguate direct client method calls first
+                    // This applies validation to direct SDK calls (including checking the required arguments are passed by keyword, not positionally)
+                    let filtered_and_mapped =
+                        method_disambiguator.disambiguate_method_calls(method_calls.clone());
+                    *method_calls = filtered_and_mapped;
+
+                    // STEP 2: Add resource direct calls (which can have positional params)
+                    // These are added AFTER disambiguation so they bypass the positional param check
                     let resource_extractor = ResourceDirectCallsExtractor::new(service_index);
                     let resource_calls = resource_extractor.extract_resource_method_calls(ast);
                     method_calls.extend(resource_calls);
 
-                    // Add waiters to extracted methods using the service model index directly
+                    // STEP 3: Add waiters and paginators (also added after disambiguation)
                     let waiters_extractor = WaitersExtractor::new(service_index);
                     let waiter_calls =
                         waiters_extractor.extract_waiter_method_calls(ast, service_index);
                     method_calls.extend(waiter_calls);
 
-                    // Add paginators to extracted methods using the service model index directly
                     let paginator_extractor = PaginatorExtractor::new(service_index);
                     let paginator_calls = paginator_extractor.extract_paginate_method_calls(ast);
                     method_calls.extend(paginator_calls);
-
-                    // Clone the method calls to pass to disambiguate_method_calls
-                    let filtered_and_mapped =
-                        method_disambiguator.disambiguate_method_calls(method_calls.clone());
-                    // Replace the method calls in place
-                    *method_calls = filtered_and_mapped;
                 }
                 ExtractorResult::Go(_, _, _) => {
                     // This shouldn't happen in Python extractor, but handle gracefully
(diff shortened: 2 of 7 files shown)


iam-policy-autopilot-policy-generation/src/extraction/python/common/argument_extractor.rs
```diff
@@ -19,7 +63,14 @@ impl ArgumentExtractor {
         let mut parameter_position = 0; // Track actual parameter position (excluding commas)
 
         for arg_node in args_nodes {
+            // Filter out comment nodes - ast-grep with relaxed strictness may capture comment nodes
+            if arg_node.kind() == "comment" {
```
mschlaipfer, line 23: Can you share a program where this is not enough?

I would strongly prefer us not to add custom parsing logic like in `strip_python_comment`, and instead traverse the AST correctly. It sees like tree-sitter does produce comment nodes.

iam-policy-autopilot-policy-generation/src/extraction/python/common/argument_extractor.rs
```diff
@@ -19,7 +63,14 @@ impl ArgumentExtractor {
...
+                continue;
+            }
+
             let arg_text = arg_node.text().to_string();
+            // Strip inline comments before processing
+            let arg_text = Self::strip_python_comment(&arg_text);
```
mschlaipfer: Let's make this change a separate PR, seems unrelated to fixing positional parameter handling

iam-policy-autopilot-policy-generation/tests/public_api_integration_test.rs
```diff
@@ -522,3 +522,195 @@ def start_policy_generation():
...
+    );
+    assert!(
+        method_names.iter().any(|n| n == "delete_bucket"),
+        "Should extract delete_bucket (from bucket.delete())"
+    );
+    // upload_file expands to operations like put_object
+    assert!(
+        method_names
+            .iter()
+            .any(|n| n.contains("put") || n.contains("upload")),
```
mschlaipfer: Can you make this exact and not a `contains` check?

iam-policy-autopilot-policy-generation/tests/public_api_integration_test.rs
```diff
@@ -522,3 +522,195 @@ def start_policy_generation():
...
+    method_names.dedup();
+
+    println!("\nExtracted method names:");
+    for name in &method_names {
+        println!("  - {}", name);
+    }
+
+    // Verify POSITIVE cases are present
+    assert!(
+        method_names.iter().any(|n| n == "list_buckets"),
```
mschlaipfer: Instead of `any` checks, can we just check against the exact list of expected names that are extracted?

iam-policy-autopilot-policy-generation/tests/public_api_integration_test.rs
```diff
@@ -522,3 +522,195 @@ def start_policy_generation():
...
+        .extract_sdk_method_calls(Language::Python, vec![source_file])
+        .await
+        .expect("Extraction should succeed");
+
+    println!("\n=== Comprehensive Parameter Validation Test ===");
+    println!("Total methods extracted: {}", extracted.methods.len());
+
+    // Serialize to inspect what was extracted
+    let methods_json =
+        serde_json::to_string_pretty(&extracted.methods).expect("Should serialize methods");
```
mschlaipfer: Let's remove the serialization and deserialization in line 643.

iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
```diff
@@ -101,27 +104,27 @@ impl Extractor for PythonExtractor {
...
-                    // Add paginators to extracted methods using the service model index directly
                     let paginator_extractor = PaginatorExtractor::new(service_index);
                     let paginator_calls = paginator_extractor.extract_paginate_method_calls(ast);
                     method_calls.extend(paginator_calls);
-
-                    // Clone the method calls to pass to disambiguate_method_calls
-                    let filtered_and_mapped =
-                        method_disambiguator.disambiguate_method_calls(method_calls.clone());
-                    // Replace the method calls in place
-                    *method_calls = filtered_and_mapped;
```
mschlaipfer, line 124: Why do we not disambiguate extracted resource calls, waiter calls and paginator calls, anymore?

iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
```diff
@@ -82,7 +84,8 @@ impl Extractor for PythonExtractor {
         let pattern = "$OBJ.$METHOD($$$ARGS)";
 
         // Find all method calls with attribute access: obj.method(args)
-        for node_match in root.find_all(pattern) {
+        let pattern_obj = Pattern::new(pattern, Python).with_strictness(MatchStrictness::Relaxed);
```
mschlaipfer: I saw `comment` nodes when parsing Python with inline comments in the ast-grep playground. I would like to understand why this is needed.

iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
afilieri-aws: Those three categories generate synthetic method calls for the sdk operations that may be called by paginators, waiters, and resource objects, so the disambiguation is redundant. 

On the other hand, if we apply disambiguation only to direct client method calls, we can discard calls with positional parameters there, since these calls only accept keyword arguments.

iam-policy-autopilot-policy-generation/src/extraction/python/common/argument_extractor.rs
afilieri-aws: As it turns out, no :) In Python, the default strictness (`smart`) can already skip comments as unnamed nodes in the pattern matching and such nodes are kept in the result with `kind() == "comment"`. Python does not allow for multiline comments, like JS, which may instead require `relaxed` strictness ([ast-grep docs](https://ast-grep.github.io/advanced/match-algorithm.html)).

So keeping this single check is enough for Python, no need for the more complex logic to strip comments also from text nodes. I cleaned it up. Thanks!

iam-policy-autopilot-policy-generation/tests/public_api_integration_test.rs
```diff
@@ -522,3 +522,187 @@ def start_policy_generation():
...
+        .expect("Extraction should succeed");
+
+    println!("\n=== Comprehensive Parameter Validation Test ===");
+    println!("Total methods extracted: {}", extracted.methods.len());
+
+    // Convert to JSON values to inspect what was extracted
+    let methods: Vec<serde_json::Value> = extracted
+        .methods
+        .iter()
+        .map(|m| serde_json::to_value(m).expect("Should serialize method"))
```
mschlaipfer, line 626: Why do we need to convert to value? Is there a visibility issue that we cannot extract the names directly from `extracted.methods` so the LLM keeps going via the `Serialize` detour??

afilieri-aws: It's a visibility issue. The fields of the extracted methods are not visible in this crate, so the integration test cannot access them directly, hence the `serde` steps in this file.

PR #34, opened 2025-11-26
fix: Filter inline comments

afilieri-aws:
*Issue #, if available:*

*Description of changes:*

Filter out comments inlined between method call parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,147 +1,321 @@
 # IAM Policy Autopilot
 
-IAM Policy Autopilot is an MCP server that helps AI coding assistants analyze application code and quickly create baseline identity-based policies that enable your application to work from day one, giving you a starting point to refine as your application evolves.
+An open source Model Context Protocol (MCP) server and command-line tool that helps your AI coding assistants quickly create baseline IAM policies that you can refine as your application evolves, so you can build faster. IAM Policy Autopilot analyzes your application code locally to generate identity-based policies for application roles, enabling faster IAM policy creation and reducing access troubleshooting time. IAM Policy Autopilot supports applications built in Python, Go, and TypeScript. 
+
+## Table of Contents
+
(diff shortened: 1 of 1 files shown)


iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
```diff
@@ -160,4 +161,61 @@ mod tests {
...
+        assert_eq!(result.method_calls_ref().len(), 1);
+
+        let method_call = &result.method_calls_ref()[0];
+        assert_eq!(method_call.name, "put_metric_alarm");
+
+        // Verify the metadata contains the correct number of parameters
+        // Comments should NOT be counted as parameters
+        if let Some(metadata) = &method_call.metadata {
+            let param_count = metadata.parameters.len();
+            // Should have 5 parameters, not 7 (which would include the 2 comments)
```
mschlaipfer: There are 3 comments

mschlaipfer: not trying to be nit-picky, just making sure we understand the change: are we getting 8 or 7 without the check this PR adds? Do we need to handle
```
# This comment should be ignored too
    # Another comment
```
different from `# This is a comment that should be ignored`?

Does the former get merged into a single AST node, so it is two comments?

afilieri-aws: I made the comment more clear: there are two lines that are only comments and one comment appended after a parameter value. They should all be stripped

afilieri-aws: Np. Without the PR, the extractor would interpret the line comments in the snippet as positional arguments, which are not acceptable for this method call (positional args are only acceptable for utility methods). 
The latest revision of the test also makes sure that comments inlined after a value of the keyword (and before the comma) are properly discarded.

PR #35, opened 2025-11-26
fix: add cache expiry logic for in-memory cache as well

weibenz1:
*Issue #, if available:*

*Description of changes:*

1. This PR introduces the expiry logic for the in-memory cache.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
@@ -22,7 +22,7 @@ type OperationName = String;
 const IAM_POLICY_AUTOPILOT: &str = "IAMPolicyAutopilot";
 // Cache files for 6 hours.
 // We can allow cache duration override in future.
-const DEFAULT_FILE_SYSTEM_CACHE_DURATION_IN_SECONDS: u64 = 21600;
+const DEFAULT_CACHE_DURATION_IN_SECONDS: u64 = 21600;
 pub(crate) type OperationToAuthorizedActionMap = HashMap<OperationName, Operation>;
 /// Service Reference data structure
 ///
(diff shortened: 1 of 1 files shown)


iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -411,19 +411,23 @@ impl RemoteServiceReferenceLoader {
         &self,
         service_name: &str,
     ) -> crate::errors::Result<Option<ServiceReference>> {
-        if let Some(cached) = self.service_cache.read().await.get(service_name) {
-            return Ok(Some(cached.clone()));
+        if let Some((cached, timestamp)) = self.service_cache.read().await.get(service_name) {
+            if let Ok(elapsed) = SystemTime::now().duration_since(*timestamp) {
+                if elapsed < Duration::from_secs(DEFAULT_FILE_SYSTEM_CACHE_DURATION_IN_SECONDS) {
```
mschlaipfer: Nit: change name of constant or introduce a separate one?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -540,12 +544,39 @@ mod tests {
...
+        // Load again - should fetch fresh data, not use expired cache
+        let result = loader.load("s3").await;
+        assert!(result.is_ok());
+
+        // Verify cache has fresh timestamp
+        let cached = loader.service_cache.read().await.get("s3").cloned();
+        assert!(cached.is_some());
+        let (_, timestamp) = cached.unwrap();
+        let elapsed = SystemTime::now().duration_since(timestamp).unwrap();
+        assert!(elapsed < Duration::from_secs(10));
```
mschlaipfer, line 576: Wouldn't this be true even if we didn't set the expired_time? The point seems to be to distinguish the loads on line 558 and 569, but both would be within that duration?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
weibenz1: good catch

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
weibenz1: 558 has a fresh timestamp.  
but after we set the expired time, the call at line 569 should make a new request and reset the cache, so we expect 576 to be fresh rather than the expired_time, hence the assertion

mschlaipfer: Ah, got it, thanks, it's late

PR #36, opened 2025-11-26
fix: rename generate-policy to generate-policies

mschlaipfer:
*Issue #, if available:*

*Description of changes:*

`generate-policy` -> `generate-policies`


CONTRIBUTING.md
```diff
@@ -64,7 +64,7 @@ We follow [Conventional Commits](https://www.conventionalcommits.org/) for commi
 For breaking changes, append `!` after the type or add `BREAKING CHANGE:` in the footer:
```
weibenz1, line 64: need to update the workflow steps; otherwise release will be blocked

mschlaipfer: Thanks both of you, great catch. I didn't include `*.yml` in my search for occurrences, now searched everywhere.

PR #47, opened 2025-12-02
test: improve configuration deserialization testing

mschlaipfer:
*Issue #, if available:*

*Description of changes:* Improve testing when configuration files have invalid data. Instead of continueing on wrong data at runtime, we now fail, as we know that we will have parsed the same files (the same way!) during testing.


iam-policy-autopilot-policy-generation/tests/resources/invalid_configs/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+This directory contains intentionally malformed configuration files used for negative testing.
+
+## Purpose
+
+These files are used to verify that the configuration parsing logic correctly rejects invalid inputs during unit tests, ensuring that errors are caught during development rather than at runtime.
+
+## How It Works
+
+The negative tests use **RustEmbed** to embed these invalid test files at compile time, just like the production code embeds valid configuration files. The key difference is:
```
weibenz1: Are we actually compiling it and do we need to?

mschlaipfer: I think `RustEmbed` actually does not embed them for debug builds, but it abstracts that away from us. I think the point is more that we use the same interface to consume these files. I can update the comment as part of fixing the build.

iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
```diff
@@ -100,22 +100,22 @@ impl Extractor for PythonExtractor {
         service_index: &ServiceModelIndex,
     ) {
         let method_disambiguator = MethodDisambiguator::new(service_index);
+        let resource_extractor = ResourceDirectCallsExtractor::new(service_index);
+        let waiters_extractor = WaitersExtractor::new(service_index);
+        let paginator_extractor = PaginatorExtractor::new(service_index);
```
adpaco, line 105: Nice.

weibenz1:
> I like the fail-fast approach for loading config files but this is a behavioral change (not just testing) and as such we should update the PR title and changelog to note this.

This should not be a breaking change to customers tho, but just to catch failures during development. All the things passed all the validation and is embedded into the final binary will remain functioning

PR #48, opened 2025-12-02
chore: Add IamPolicyAutopilot policy ID to access denied flows

LukeKennedy:
*Description of changes:*
The access denied flows only place the IamPolicyAutopilot tag in the statement ID, not the policy ID. This change adds it to the policy ID. 

This is `Option<String>` because there is no assurance the incoming policy when merging has the field present, so it must be able to be optionally included for proper parsing. I've manually tested it with a few policies and screwed up others to make sure it handles the merging gracefully.

### DIFF
--- a/iam-policy-autopilot-access-denied/src/commands/apply.rs
+++ b/iam-policy-autopilot-access-denied/src/commands/apply.rs
@@ -1,7 +1,7 @@
 //! Apply logic for IAM Policy Autopilot service
 
 use crate::aws::iam_client::{find_canonical_policy, put_inline_policy};
-use crate::aws::policy_naming::{build_canonical_policy_name, build_statement_sid};
+use crate::aws::policy_naming::{build_canonical_policy_name, build_statement_sid, POLICY_PREFIX};
 use crate::aws::principal::resolve_principal;
 use crate::aws::sts::caller_account_id;
 use crate::synthesis::build_single_statement;
@@ -82,6 +82,10 @@ impl super::service::IamPolicyAutopilotService {
             sort_statements(&mut merged_statements);
 
             let policy_doc = crate::types::PolicyDocument {
+                id: existing
+                    .document
+                    .id
+                    .or_else(|| Some(POLICY_PREFIX.to_string())),
                 version: "2012-10-17".to_string(),
                 statement: merged_statements,
             };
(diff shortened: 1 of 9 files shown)


iam-policy-autopilot-access-denied/src/commands/apply.rs
```diff
@@ -82,6 +82,7 @@ impl super::service::IamPolicyAutopilotService {
             sort_statements(&mut merged_statements);
 
             let policy_doc = crate::types::PolicyDocument {
+                id: Some("IamPolicyAutopilot".to_string()),
```
mschlaipfer: For an existing policy, are we currently dropping the `Id` if there was one set? Now we would replace it with `IamPolicyAutopilot`? Do we need to check if the existing policy has an `Id` set?

iam-policy-autopilot-access-denied/src/types.rs
```diff
@@ -42,6 +42,7 @@ impl ParsedDenial {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "PascalCase")]
 pub struct Statement {
+    #[serde(default, skip_serializing_if = "String::is_empty")]
```
karanjitsingh: `skip_serializing_if = "String::is_empty"` makes it quite sneaky why not `Option<String>`

LukeKennedy: Honestly, I don't remember adding this. Will revert since I'm not touching anything SID related.

iam-policy-autopilot-mcp-server/src/tools/generate_policy_for_access_denied.rs
```diff
@@ -98,6 +99,7 @@ mod tests {
         };
 
         let sample_policy = PolicyDocument {
+            id: Some("IamPolicyAutopilot".to_string()),
```
weibenz1: probably should reuse this constant - https://github.com/awslabs/iam-policy-autopilot/blob/fa61060532a763879168114b75dd01d578609db4/iam-policy-autopilot-access-denied/src/aws/policy_naming.rs#L9, or create a new one for PolicyID

PR #50, opened 2025-12-03
fix: improve Go SDK parameter extraction and service disambiguation

adpaco-aws:
*Description of changes:* Fixes two critical bugs in Go SDK method extraction that were causing incorrect service disambiguation and filtering of valid operations.

#### Incomplete struct field extraction

The extractor was using string parsing to extract field names from Go struct literals, which failed to reliably extract parameters. This prevented proper service disambiguation.

Fix: Implemented AST-based field extraction to accurately parse struct literals and extract field names directly from the syntax tree.

#### Case-sensitive parameter validation

AWS service models use inconsistent casing - some services use PascalCase while others use camelCase. Since the Go SDK always uses PascalCase, case-sensitive validation was incorrectly rejecting valid operations or matching them to the wrong services.

Fix: Changed parameter validation to use case-insensitive comparison for both valid parameter checks and required parameter checks.

### Testing

Added 5 new tests covering:
 - Missing required parameters
 - Required parameter validation
 - Variable-based parameters
 - Service disambiguation by required parameters
 - Nested map keys in struct literals

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/extraction/go/disambiguation.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/go/disambiguation.rs
@@ -241,6 +241,8 @@ impl<'a> GoMethodDisambiguator<'a> {
 
         // Get all valid parameters from the shape (lowercase for case-insensitive comparison)
         // AWS models are inconsistent - some use PascalCase, some use camelCase
+        // TODO: Canonicalize casing during deserialization instead of at comparison time
+        // See: https://github.com/awslabs/iam-policy-autopilot/issues/57
         let valid_params_lower: HashSet<String> =
             shape.members.keys().map(|k| k.to_lowercase()).collect();
         log::debug!(
--- a/iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
@@ -3,6 +3,7 @@
 use crate::extraction::extractor::{Extractor, ExtractorResult};
 use crate::extraction::go::disambiguation::GoMethodDisambiguator;
 use crate::extraction::go::features_extractor::GoFeaturesExtractor;
+use crate::extraction::go::node_kinds;
 use crate::extraction::go::paginator_extractor::GoPaginatorExtractor;
 use crate::extraction::go::types::{GoImportInfo, ImportInfo};
 use crate::extraction::go::waiter_extractor::GoWaiterExtractor;
@@ -167,7 +168,9 @@ rule:
                 .filter(|child| {
                     // Filter out parentheses and commas, keep only actual argument nodes
                     let kind = child.kind();
-                    kind != "(" && kind != ")" && kind != ","
+                    kind != node_kinds::LEFT_PAREN
+                        && kind != node_kinds::RIGHT_PAREN
+                        && kind != node_kinds::COMMA
                 })
                 .collect();
 
@@ -279,7 +282,7 @@ rule:
         );
 
         // Check if this node is directly a composite_literal
-        if node.kind() == "composite_literal" {
+        if node.kind() == node_kinds::COMPOSITE_LITERAL {
             log::debug!("Node is directly a composite_literal");
             return self.extract_fields_from_composite_literal(node);
         }
@@ -287,16 +290,16 @@ rule:
         // Check immediate children for unary_expression (for &Type{...} pattern)
         for child in node.children() {
             log::debug!("Checking child node: kind={}", child.kind());
-            if child.kind() == "unary_expression" {
+            if child.kind() == node_kinds::UNARY_EXPRESSION {
                 // Check the unary_expression's children for composite_literal
                 for unary_child in child.children() {
                     log::debug!("Checking unary child: kind={}", unary_child.kind());
-                    if unary_child.kind() == "composite_literal" {
+                    if unary_child.kind() == node_kinds::COMPOSITE_LITERAL {
                         log::debug!("Found composite_literal under unary_expression");
                         return self.extract_fields_from_composite_literal(&unary_child);
                     }
                 }
-            } else if child.kind() == "composite_literal" {
+            } else if child.kind() == node_kinds::COMPOSITE_LITERAL {
                 log::debug!("Found composite_literal as direct child");
                 return self.extract_fields_from_composite_literal(&child);
             }
@@ -352,14 +355,14 @@ rule:
         // A composite_literal has exactly one literal_value child in the AST.
         let literal_value = composite_literal
             .children()
-            .find(|child| child.kind() == "literal_value");
+            .find(|child| child.kind() == node_kinds::LITERAL_VALUE);
 
         if let Some(literal_value) = literal_value {
             log::debug!("Found literal_value, extracting keyed_elements...");
             // Extract field names from keyed_element nodes
             for element in literal_value.children() {
                 log::debug!("Literal_value child: kind={}", element.kind());
-                if element.kind() == "keyed_element" {
+                if element.kind() == node_kinds::KEYED_ELEMENT {
                     // Get the first child, which is the field name (literal_element)
                     if let Some(field_name_node) = element.children().next() {
                         log::debug!(
@@ -367,7 +370,7 @@ rule:
                             field_name_node.kind(),
                             field_name_node.text()
                         );
-                        if field_name_node.kind() == "literal_element" {
+                        if field_name_node.kind() == node_kinds::LITERAL_ELEMENT {
                             field_names.push(field_name_node.text().to_string());
                         }
                     }
--- a/iam-policy-autopilot-policy-generation/src/extraction/python/common/argument_extractor.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/python/common/argument_extractor.rs
@@ -4,6 +4,7 @@
 //! from ast-grep nodes, handling keyword arguments, positional arguments,
 //! and dictionary unpacking consistently across all Python extractors.
 
+use crate::extraction::python::node_kinds;
 use crate::extraction::{Parameter, ParameterValue};
 use ast_grep_language::Python;
 
@@ -20,7 +21,7 @@ impl ArgumentExtractor {
 
         for arg_node in args_nodes {
             // Filter out comment nodes
-            if arg_node.kind() == "comment" {
+            if arg_node.kind() == node_kinds::COMMENT {
                 continue;
             }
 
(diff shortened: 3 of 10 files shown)


iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
```diff
@@ -156,9 +156,35 @@ rule:
             return None;
         };
 
-        // Extract arguments - get_multiple_matches returns Vec<Node> directly
-        let args_nodes = env.get_multiple_matches("ARGS");
-        let arguments = self.extract_arguments(&args_nodes);
+        // Extract arguments - ARGS captures the entire argument_list node
+        // We need to get its children to access individual arguments
+        let arguments = if let Some(args_node) = env.get_match("ARGS") {
```
mschlaipfer, line 162: Did you try to get `get_multiple_matches` (possibly with a different ast-grep query) to work? When I initially looked into this, I also did this AST walk, but was hoping we don't have to.

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
```diff
@@ -196,19 +223,127 @@ rule:
...
         parameters
     }
 
+    /// Extract type name from struct literal text
+    fn extract_type_from_struct_literal(&self, text: &str) -> Option<String> {
+        let trimmed = text.trim();
+        if trimmed.starts_with('&') {
+            if let Some(brace_pos) = trimmed.find('{') {
+                let type_part = trimmed[1..brace_pos].trim();
+                return Some(format!("*{}", type_part));
```
mschlaipfer: Do you think we need the `*` prefix?

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
```diff
@@ -196,19 +223,127 @@ rule:
...
+                return self.extract_fields_from_composite_literal(&child);
+            }
+        }
+
+        log::debug!("No composite_literal found");
+        field_names
+    }
+
+    /// Extract field names from a composite_literal node
+    fn extract_fields_from_composite_literal(
```
mschlaipfer, line 346: Could you add an example for what's a `composite_literal`?

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
```diff
@@ -196,19 +223,127 @@ rule:
...
+                                first_child.kind(),
+                                first_child.text()
+                            );
+                            if first_child.kind() == "literal_element" {
+                                field_names.push(first_child.text().to_string());
+                            }
+                        }
+                    }
+                }
+                break; // Only process the first literal_value
```
mschlaipfer: Could you explain this?

iam-policy-autopilot-policy-generation/src/extraction/python/common/argument_extractor.rs
```diff
@@ -52,6 +52,7 @@ impl ArgumentExtractor {
                     value: Self::extract_parameter_value(&arg_text),
                     position: parameter_position,
                     type_annotation: None,
+                    struct_fields: None,
```
mschlaipfer, line 56: Could we implement `Default` and use `..Default::default()` here, so we don't need to do these updates across the code base in the future?

If you agree, could you open an issue?

iam-policy-autopilot-policy-generation/src/extraction/go/disambiguation.rs
```diff
@@ -215,8 +223,13 @@ impl<'a> GoMethodDisambiguator<'a> {
...
+                    // Use AST-extracted fields if available (only top-level fields, no nested map keys)
+                    if let Some(fields) = struct_fields {
+                        return Some(fields.clone());
+                    }
+
+                    // Fallback to string parsing for backwards compatibility
                     let value_str = value.as_string();
-                    // Extract field names from struct literals
                     if value_str.contains('{') && value_str.contains('}') {
                         Some(self.extract_struct_field_names(value_str))
```
mschlaipfer: Do we really need this? Can this ever work?

iam-policy-autopilot-policy-generation/src/extraction/go/disambiguation.rs
```diff
@@ -228,30 +241,66 @@ impl<'a> GoMethodDisambiguator<'a> {
...
-        let _required_params: HashSet<String> = shape
+        let required_params: HashSet<String> = shape
             .required
             .as_ref()
             .map(|req| req.iter().cloned().collect())
             .unwrap_or_default();
+        log::debug!("Required parameters from AWS model: {:?}", required_params);
 
         // Get all valid parameters from the shape
         let valid_params: HashSet<String> = shape.members.keys().cloned().collect();
```
weibenz1: better to do the lowercase conversion here for once and all?

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
```diff
@@ -196,19 +223,127 @@ rule:
...
+                    log::debug!("Literal_value child: kind={}", element.kind());
+                    if element.kind() == "keyed_element" {
+                        // Get the first child which should be the field name
+                        if let Some(first_child) = element.children().next() {
+                            log::debug!(
+                                "Keyed_element first child: kind={}, text={:?}",
+                                first_child.kind(),
+                                first_child.text()
+                            );
+                            if first_child.kind() == "literal_element" {
```
weibenz1: can we define an enum for all the types/kinds of the nodes?

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
adpaco: I tried to but the `$$ARGS` pattern returns a single node with children nodes for parentheses and arguments in the Go AST-based approach, which doesn't align entirely with what you get in Python with the string-based approach. Maybe there is a way but I wonder if we should be looking at how Python does this instead.

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
adpaco: Not really, removed. If we need to distinguish reference types in the future we can consider this or other options.

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
adpaco: Added docs and examples.

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
adpaco: That was unnecessary, there shouldn't be any other children (`literal_value`s) at that level. Anyway, updated this to avoid the loop structure.

iam-policy-autopilot-policy-generation/src/extraction/go/disambiguation.rs
adpaco: We don't, removed.

iam-policy-autopilot-policy-generation/src/extraction/go/disambiguation.rs
adpaco: Thanks, updated the code to do it once.

iam-policy-autopilot-policy-generation/src/extraction/python/common/argument_extractor.rs
adpaco: I do agree it'd be cleaner but this is an enum variant so we can't implement `Default` on it as is. We could encapsulate the fields of `Parameter::Positional` into a struct and implement `Default` on it, let me know if you'd like that.

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
adpaco: No, these values are based on the [Go Tree-sitter grammar](https://github.com/tree-sitter/tree-sitter-go/blob/master/grammar.js). These grammars are language-dependent and dynamic by design (i.e., they can be updated) so we cannot use an enum. However, I could define constants for the node kinds we use in this code if you're worried about typos. What do you think?

iam-policy-autopilot-policy-generation/src/extraction/python/common/argument_extractor.rs
mschlaipfer: I think we should do that, but we can do it as part of a dedicated refactoring which introduces `Default` more broadly.

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
mschlaipfer: Introducing constants for that reason seems sensible, but we can do it in a separate PR

iam-policy-autopilot-policy-generation/src/extraction/go/disambiguation.rs
```diff
@@ -215,69 +221,72 @@ impl<'a> GoMethodDisambiguator<'a> {
...
             .required
             .as_ref()
             .map(|req| req.iter().cloned().collect())
             .unwrap_or_default();
+        log::debug!("Required parameters from AWS model: {:?}", required_params);
+
+        // Get all valid parameters from the shape (lowercase for case-insensitive comparison)
+        // AWS models are inconsistent - some use PascalCase, some use camelCase
+        let valid_params_lower: HashSet<String> =
+            shape.members.keys().map(|k| k.to_lowercase()).collect();
```
mschlaipfer, line 247: We need to address this more broadly for other languages as well. We should canonicalize during deserialization. Created https://github.com/awslabs/iam-policy-autopilot/issues/57

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
mschlaipfer: Thinking about this some more: I tried to see if using the constants over string literals could be enforced using a lint, but I don't see such a lint. Should we still do it and ensure that the constants are used in code review?

mschlaipfer (dismissed):

iam-policy-autopilot-policy-generation/src/extraction/python/common/argument_extractor.rs
adpaco: Opened #61 and referenced it in the code.

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
adpaco: We can write a custom lint for this. Opened #60 and referenced it.

iam-policy-autopilot-policy-generation/src/extraction/go/disambiguation.rs
adpaco: Agree, linked in the code.

iam-policy-autopilot-policy-generation/src/extraction/go/extractor.rs
weibenz1: thanks! constants also make sense. I was just worried about typos plus long term maintainability of our codebase

PR #54, opened 2025-12-04
ci: address security findings

weibenz1:
*Issue #, if available:*

*Description of changes:*

Fix security warnings in - https://github.com/awslabs/iam-policy-autopilot/security/code-scanning.

### DIFF
--- a/.github/workflows/build_and_publish.yml
+++ b/.github/workflows/build_and_publish.yml
@@ -4,9 +4,14 @@ on:
   release:
     types: [published]
 
+permissions:
+  contents: read
+
 jobs:
   build_wheels:
     name: Build ${{ matrix.target }} wheel
+    permissions:
+      contents: read
     # Define the matrix for runners and Rust targets
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


.github/workflows/slack_notifications.yml
```diff
@@ -10,14 +10,17 @@ on:
...
+  contents: read
+
 jobs:
   notify:
     runs-on: ubuntu-latest
     permissions:
       contents: read
     steps:
       - name: Send Event Notification
-        if: ${{ secrets.SLACK_WEBHOOK_URL != '' }}
```
weibenz1, line 20: fixing syntax error.
continue-on-error will make this workflow pass regardless of whether notification was actually sent; don't want to report a failed status for a notification workflow

.github/workflows/build_and_publish.yml
```diff
@@ -242,6 +249,7 @@ jobs:
     runs-on: ubuntu-latest
     environment: Release
     permissions:
+      contents: read
```
karanjitsingh, line 252: just confirming, not everyone can start this workflow right?

weibenz1: this defines what permission the github action has on the repo; the trigger is defined at the top. 

Manual trigger can only be done if there's a `on: workflow_dispatch:`

PR #55, opened 2025-12-04
refactor: SDK model loading

mschlaipfer:
*Issue #, if available:*

*Description of changes:* Some cleanup in data loading: Moving to a setup, where the `...Raw` structs have the raw embedded data, and the `...Data` structs provide views on top of raw data. I think this separation makes sense, as we sometimes combine data from multiple files, and consumers of the data should not need to know which file it comes from.
This also gets us a step closer to ensuring we load data into memory only once.

Also removed `WaitersRegistry` and used `serde_json` directly instead of the `JsonProvider`. If we decide to add a Wasm compilation target, we will need to revisit this.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/embedded_data.rs
+++ b/iam-policy-autopilot-policy-generation/src/embedded_data.rs
@@ -5,9 +5,11 @@
 //! have been simplified to remove documentation and examples, reducing binary size
 //! while maintaining all essential functionality.
 
+use std::borrow::Cow;
+use std::collections::HashMap;
+
 use crate::errors::{ExtractorError, Result};
 use crate::extraction::sdk_model::SdkServiceDefinition;
-use crate::providers::JsonProvider;
 use rust_embed::RustEmbed;
 
 /// Embedded AWS service definitions with compression
@@ -18,7 +20,7 @@ use rust_embed::RustEmbed;
 #[derive(RustEmbed)]
 #[folder = "target/botocore-data-simplified"]
 #[include = "*.json"]
-pub struct Botocore;
+struct BotocoreRaw;
 
 /// Embedded AWS boto3 resource definitions
 ///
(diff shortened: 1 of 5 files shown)


iam-policy-autopilot-policy-generation/src/embedded_data.rs
```diff
@@ -232,28 +210,31 @@ impl EmbeddedBoto3Data {
...
+        Boto3UtilitiesRaw::get_utilities_mapping()
     }
 }
 
 /// Embedded AWS service data manager
 ///
 /// Provides convenient access to embedded AWS service definitions with
 /// automatic decompression and JSON parsing.
-pub(crate) struct EmbeddedServiceData;
+pub(crate) struct BotocoreData;
```
weibenz1, line 235: this separation makes sense

iam-policy-autopilot-policy-generation/src/extraction/javascript/shared.rs
```diff
@@ -5,9 +5,26 @@
...
+/// that defines how lib-* submodule commands map to client-* commands.
+#[derive(RustEmbed)]
+#[folder = "resources/config/sdks"]
+#[include = "js_v3_libraries.json"]
+struct JsV3Libraries;
+
+impl JsV3Libraries {
+    /// Get the JavaScript SDK v3 libraries mapping configuration
+    fn get_libraries_mapping() -> Option<std::borrow::Cow<'static, [u8]>> {
+        Self::get("js_v3_libraries.json").map(|file| file.data)
```
weibenz1, line 24: I believe we already have tests to capture changes to this file (e.g. file not found + format changes?)

weibenz1 (approved):

iam-policy-autopilot-policy-generation/src/extraction/javascript/shared.rs
mschlaipfer: I will update https://github.com/awslabs/iam-policy-autopilot/pull/47 to test this as well.

PR #62, opened 2025-12-09
feat: add custom dylint lint to enforce node_kinds constants usage

### LINKED ISSUE #60: Enforce using constants for node kind comparison using custom lint
We want to avoid comparing node kinds to string so we're going to start using constants instead. It should be possible to enforce it using a custom clippy lint.

See original discussion:

_Originally posted by @mschlaipfer in https://github.com/awslabs/iam-policy-autopilot/pull/50#discussion_r2592245363_

adpaco-aws:
*Issue #, if available:* Resolves #60

*Description of changes:* Adds a custom dylint lint that enforces the use of `node_kinds` constants instead of string literals when comparing Tree-sitter node types. The lint detects any pattern where `.kind()` is compared with a string literal (e.g., `node.kind() == "composite_literal"`) and suggests defining a constant in a `node_kinds` module.

This is enforced automatically in CI via a new `custom-lints` job that runs on every PR, preventing regressions from the recent node kinds constants refactor.

### DIFF
--- a/iam-policy-autopilot-lints/src/node_kind_literal.rs
+++ b/iam-policy-autopilot-lints/src/node_kind_literal.rs
@@ -34,7 +34,7 @@ dylint_linting::declare_late_lint! {
     /// }
     /// ```
     pub NODE_KIND_LITERAL,
-    Warn,
+    Deny,
     "use of string literals in comparisons with .kind() method calls"
 }
 
--- a/iam-policy-autopilot-lints/ui/main.stderr
+++ b/iam-policy-autopilot-lints/ui/main.stderr
@@ -1,13 +1,13 @@
-warning: comparing .kind() with string literal "composite_literal"
+error: comparing .kind() with string literal "composite_literal"
   --> $DIR/main.rs:15:23
    |
 LL |     if node.kind() == "composite_literal" {
    |                       ^^^^^^^^^^^^^^^^^^^
    |
    = help: define and use a constant like `const COMPOSITE_LITERAL: &str = "composite_literal";` in a node_kinds module
-   = note: `#[warn(node_kind_literal)]` on by default
+   = note: `#[deny(node_kind_literal)]` on by default
 
-warning: comparing .kind() with string literal "unary_expression"
+error: comparing .kind() with string literal "unary_expression"
   --> $DIR/main.rs:20:8
    |
 LL |     if "unary_expression" == node.kind() {
@@ -15,7 +15,7 @@ LL |     if "unary_expression" == node.kind() {
    |
    = help: define and use a constant like `const UNARY_EXPRESSION: &str = "unary_expression";` in a node_kinds module
 
-warning: comparing .kind() with string literal "literal_value"
+error: comparing .kind() with string literal "literal_value"
   --> $DIR/main.rs:25:23
    |
 LL |     if node.kind() != "literal_value" {
@@ -23,7 +23,7 @@ LL |     if node.kind() != "literal_value" {
    |
    = help: define and use a constant like `const LITERAL_VALUE: &str = "literal_value";` in a node_kinds module
 
-warning: comparing .kind() with string literal "some_new_node_type"
+error: comparing .kind() with string literal "some_new_node_type"
   --> $DIR/main.rs:30:23
    |
 LL |     if node.kind() == "some_new_node_type" {
@@ -31,5 +31,5 @@ LL |     if node.kind() == "some_new_node_type" {
    |
    = help: define and use a constant like `const SOME_NEW_NODE_TYPE: &str = "some_new_node_type";` in a node_kinds module
 
-warning: 4 warnings emitted
+error: aborting due to 4 previous errors
 
--- a/iam-policy-autopilot-policy-generation/src/extraction/go/node_kinds.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/go/node_kinds.rs
@@ -9,9 +9,6 @@
 //!
 //! Note: The actual values come from the Tree-sitter Go grammar and cannot be
 //! changed. We're just providing named constants to avoid magic strings.
-//!
-//! TODO: Add automated linting to enforce usage of these constants instead of
-//! string literals. See: https://github.com/awslabs/iam-policy-autopilot/issues/60
 
 /// A composite literal node (e.g., `Type{field: value}`)
 pub(crate) const COMPOSITE_LITERAL: &str = "composite_literal";
--- a/iam-policy-autopilot-policy-generation/src/extraction/python/node_kinds.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/python/node_kinds.rs
@@ -9,9 +9,6 @@
 //!
 //! Note: The actual values come from the Tree-sitter Python grammar and cannot be
 //! changed. We're just providing named constants to avoid magic strings.
-//!
-//! TODO: Add automated linting to enforce usage of these constants instead of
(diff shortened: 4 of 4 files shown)


iam-policy-autopilot-lints/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "node_kind_literal"
+version = "0.1.0"
+authors = ["IAM Policy Autopilot Contributors"]
+description = "Dylint lint to enforce usage of node_kinds constants instead of string literals"
```
mschlaipfer: Let's make this generic

iam-policy-autopilot-lints/rust-toolchain
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "nightly-2025-09-18"
```
mschlaipfer, line 5: Do we need nightly? Could you add a comment why, if so

iam-policy-autopilot-lints/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+edition = "2021"
+publish = false
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+clippy_utils = { git = "https://github.com/rust-lang/rust-clippy", rev = "20ce69b9a63bcd2756cd906fe0964d1e901e042a" }
```
mschlaipfer, line 16: Is there a reason for this particular version? Doesn't seem to be the latest one. Could you add a comment. I see that this is not published as a crate.

.github/workflows/pr-checks.yml
```diff
@@ -166,3 +166,35 @@ jobs:
 
       - name: Run clippy
         run: cargo clippy --workspace -- -D warnings
+
+  custom-lints:
+    name: Custom Lints (dylint)
+    needs: commit-messages
```
mschlaipfer, line 172: I see this also on top, why is there `needs: commit-messages`? I'm not familiar with Github Actions

iam-policy-autopilot-lints/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "node_kind_literal"
```
mschlaipfer: This doesn't look right

iam-policy-autopilot-lints/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version = "0.1.0"
+authors = ["IAM Policy Autopilot Contributors"]
```
mschlaipfer: I see that we don't have authors in the workspace's `Cargo.toml`, should we add it there and infer it? Not sure that works since we explicitly exclude this crate.

iam-policy-autopilot-lints/src/node_kind_literal.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// ```
+    pub NODE_KIND_LITERAL,
+    Warn,
+    "use of string literals in comparisons with .kind() method calls"
+}
+
+/// Check if an expression is a call to the `.kind()` method
+fn is_kind_method_call(expr: &Expr<'_>) -> bool {
+    if let ExprKind::MethodCall(path_segment, _, _, _) = &expr.kind {
+        path_segment.ident.name.as_str() == "kind"
```
mschlaipfer: Do we have type information available here to make this check more specific?

mschlaipfer (changes requested):
Nice! Just some questions and minor changes.

iam-policy-autopilot-lints/Cargo.toml
adpaco: Sorry, looks like I didn't commit my latest changes here - it's generic now.

iam-policy-autopilot-lints/rust-toolchain
adpaco: Yes, we do need nightly to access the Rust compiler APIs when building the lints. Added a comment.

iam-policy-autopilot-lints/Cargo.toml
adpaco: No reason other than it builds with the nightly compiler version we're using. Added a comment about this.

.github/workflows/pr-checks.yml
adpaco: This means it depends on the `commit-messages` jobs, so it won't start this or other workflows like `clippy` until that's passed. It's much easier to visualize here: https://github.com/awslabs/iam-policy-autopilot/actions/runs/20071757630

iam-policy-autopilot-lints/Cargo.toml
adpaco: Fixed.

iam-policy-autopilot-lints/Cargo.toml
adpaco: No, that won't work. Remove the `authors` attribute.

iam-policy-autopilot-lints/src/node_kind_literal.rs
adpaco: Yes, it's available. But what additional checks would you do? Checking that this expression is a string type might not add much since we're already checking the other operands is a `StringLit` (and if this value cannot be coerced into a string type then it wouldn't typecheck).

mschlaipfer: That `kind` is called on a treesitter AST node. Not blocking on it, though.

mschlaipfer (dismissed):

iam-policy-autopilot-lints/src/node_kind_literal.rs
adpaco: Right, added it.

iam-policy-autopilot-lints/src/node_kind_literal.rs
```diff
@@ -38,10 +38,18 @@ dylint_linting::declare_late_lint! {
...
-        path_segment.ident.name.as_str() == "kind"
+/// Check if an expression is a call to the `.kind()` method on a tree-sitter Node
+fn is_kind_method_call<'tcx>(cx: &rustc_lint::LateContext<'tcx>, expr: &'tcx Expr<'_>) -> bool {
+    if let ExprKind::MethodCall(path_segment, receiver, _, _) = &expr.kind {
+        if path_segment.ident.name.as_str() != "kind" {
+            return false;
+        }
+
+        // Check if the receiver type contains "Node" in its path
+        // This catches tree_sitter::Node and similar types
```
mschlaipfer, line 49: Given that we don't have unit tests using tree_sitter, I trust you checked that this is the right thing to check for?

iam-policy-autopilot-lints/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "iam_policy_autopilot_lints"
+version = "0.1.0"
```
weibenz1, line 3: this will not need to be updated at all right?

iam-policy-autopilot-lints/src/node_kind_literal.rs
adpaco: Yes, this triggers on our code. What I realized just now is that we're only producing warnings and that won't block the CI run. So I made the `node_kind_literal` lint to `Deny` by default (couldn't find an option like `-D warnings` for `dylint`).

iam-policy-autopilot-lints/Cargo.toml
adpaco: No, the version doesn't matter especially since `publish = false`. I believe this is what `cargo new` adds by default.

Issue #64, opened 2025-12-09
Add support for Kiro Powers


weibenz1:
Kiro has launched [Power](https://kiro.dev/blog/introducing-powers/) recently, we should support it natively.

Here are the main advantages:
1. LLMs will be more accurate and proactive in terms of deciding when to trigger IPA MCP
2. Installation for IPA MCP will be easier.

PR #68, opened 2025-12-11
feat: add kiro power configuration

hwei0:
*Issue #, if available:*
https://github.com/awslabs/iam-policy-autopilot/issues/64
*Description of changes:*
Adding a kiro power configuration directory, consisting of:
1. a `POWER.md` readme, specifying onboarding information and very brief "steering" information (specifically, two example use cases)
2. a `mcp.json` config. 

(note: in our installation guide we specify two different `mcp.json` configs, one for pip+binary installation, and one for `uv`. Kiro powers don't seem to support selectively picking/enabling one `mcp.json` config out of multiple configs, so I specified in the Kiro power's onboarding doc that the agent must run `uv tool install iam-policy-autopilot` if it detects a `uv`-based installation.)

*Testing*: installed it into my Kiro IDE.

(note: Kiro offers a feature where kiro power can be automatically detected and installed from github repositories, but this functionality seems to be broken. So for now it seems that the power must be installed by doing `git clone` and installing from a local folder; I specified this as the way to go in the README).

### DIFF
--- a/iam-policy-autopilot-access-denied/src/aws/iam_client.rs
+++ b/iam-policy-autopilot-access-denied/src/aws/iam_client.rs
@@ -181,10 +181,14 @@ pub async fn find_canonical_policy(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::types::{ActionType, Statement};
+    use crate::{
+        aws::policy_naming::POLICY_PREFIX,
+        types::{ActionType, Statement},
+    };
 
     fn sample_policy() -> PolicyDocument {
         PolicyDocument {
+            id: Some(POLICY_PREFIX.to_string()),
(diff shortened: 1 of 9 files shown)


power-iam-policy-autopilot/mcp.json
```diff
@@ -0,0 +1,14 @@
+{
+  "mcpServers": {
+    "iam-policy-autopilot": {
```
hwei0: the [kiro docs](https://kiro.dev/docs/powers/create/) say this should be `iam-policy-autopilot-mcp` instead of `iam-policy-autopilot` or else the kiro power will break, but it seems this works for some reason. maybe their validation is broken. so i'll leaving it like this for now.

Issue #66, opened 2025-12-10
IPA doesn't generate permissions for SDK method calls like describe_db_clusters


mschlaipfer:
We received a report that `modify_db_cluster`, `modify_db_cluster`, and `modify_db_instance` are missed by IPA. I assume this is due to a bug in how we rename Python's camel case method names to the respective operation names which have a `DB` in their name.

PR #70, opened 2025-12-11
Use SDK info to find operation from method name

mschlaipfer:
*Issue #, if available:* https://github.com/awslabs/iam-policy-autopilot/issues/66

*Description of changes:*

We now look up method names in the service reference when we detect the SDK type is boto3 (i.e., when the input files are Python). This fixes an issue where we wrongly converted method names to operation names via `convert_case`'s `to_case(Case::PascalCase)` before, which fails for operations like `ModifyDBCluster`, where the snake_case method name is `modify_db_cluster` and the conversion results in `ModifyDbCluster`.
For other SDKs we use the extracted method names as-is as the operation name. The conversion to PascalCase, also changed `ModifyDBCluster` to `ModifyDbCluster`, leading to the same error for Go, Typescript and Javascript.

I introduced `SdkType::Boto3` to follow the data the service reference provides, such as
```
{
      "Name" : "rds",
      "Method" : "modify_db_cluster",
      "Package" : "Boto3"
}
```

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
@@ -14,7 +14,7 @@ use crate::enrichment::service_reference::ServiceReference;
 use crate::enrichment::{Condition, ServiceReferenceLoader};
 use crate::errors::{ExtractorError, Result};
 use crate::service_configuration::ServiceConfiguration;
-use crate::SdkMethodCall;
+use crate::{SdkMethodCall, SdkType};
 
 /// ResourceMatcher coordinates OperationAction maps and Service Reference data to generate enriched method calls
 ///
@@ -25,6 +25,7 @@ use crate::SdkMethodCall;
 pub(crate) struct ResourceMatcher {
     service_cfg: Arc<ServiceConfiguration>,
     fas_maps: OperationFasMaps,
+    sdk: SdkType,
 }
 
 // TODO: Make this configurable: https://github.com/awslabs/iam-policy-autopilot/issues/19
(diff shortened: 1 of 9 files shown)


iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -183,22 +189,46 @@ impl ResourceMatcher {
...
-        );
-
+        let initial = {
+            let initial_service_name = self
+                .service_cfg
+                .rename_service_service_reference(service_name);
+            // Determine the initial operation name, with special handling for Python's boto3 method names
+            let initial_operation_name = if self.sdk == SdkType::Boto3 {
+                // Try to load service reference and look up the boto3 method mapping
+                service_reference_loader
```
weibenz1, line 199: down the road this entire logic should be abstracted into `get_operation_name(method_name)` and abstracted within service reference

hence why I had `get_operation_to_authorized_action`

mschlaipfer: Sure, we can factor this out into a method to make it more testable, but I don't think we need this logic anywhere else. I removed `get_operation_to_authorized_actions` as it led to us making multiple loads for the same service (besides this logic being `get_operation_name_from_method_name` as you point out)

weibenz1: >  I removed get_operation_to_authorized_actions as it led to us making multiple loads for the same service 

the `load` automatically reads from cache so it's same as reading from map when cache hit but gives additional chance of refreshing expired cache, which is why I don't want to directly expose it via static map - we lose all the reloading/refreshing

Issue #63, opened 2025-12-09
Setup automated workflows to fetch latest boto3 changes


weibenz1:
IPA relies on boto3 data to be accurate for generating proper policies. To keep up with the update from boto3, we'd need an automated workflow that:
1. runs on a frequency
2. detect changes between head commit and new commits in submodule
3. run tests to ensure new changes will not break anything
4. create a PR back for review, along with summary of changes.

PR #74, opened 2025-12-11
feat: add dependabot submodule version sync

hwei0:
*Issue #, if available:*
https://github.com/awslabs/iam-policy-autopilot/issues/63
*Description of changes:*
add dependabot to create PRs to sync submodule versions on weekly basis

*Testing*
see https://github.com/hwei0/iam-policy-autopilot/pulls for example PRs created

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+
+updates:
+  - package-ecosystem: gitsubmodule
+    schedule:
+        interval: "weekly"
+    directory: /


weibenz1:
> everything outside of that data (say a change to a documentation string) is not relevant to IPA

I agree with this, but I also find syncing the commits for e.g. doc-only updates not harmful. Is there a specific use for knowing the exact change (maybe part of our release changelog)?

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+
+updates:
```
weibenz1, line 3: anyway to look for new tags/new releases? not sure if replicating head commit is a good practice - we might accidentally take in bad commits etc.

hwei0: This doesn't seem to be supported by dependabot natively, although they are tracking it as an unresolved issue: https://github.com/dependabot/dependabot-core/issues/1639 

I think if we want to do any customization (eg do it on triggered basis, only track specific directories in the boto3/botocore dependencies, track specific release tags) then we will need to find another 3rd party bot/action or make our own action. E.g. https://docs.renovatebot.com/modules/manager/git-submodules/ seems like a decent 3p alternative

weibenz1: yeah, action will offer more customizability. the part taking time would be to setup a bot account and let it create a PR, but there might be actions already automated this

PR #79, opened 2025-12-15
refactor: remove waiters from SdkServiceDefinition

mschlaipfer:
*Issue #, if available:*

*Description of changes:* Stores the waiter model the same as "regular" method calls by introducing `waiter_lookup` in `ServiceModelIndex`, removing the need for a two-step service name and then waiter definition lookup. Also removes some duplication in the synthetic call generation related to waiters by introducing `create_synthetic_call_internal`.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
@@ -8,7 +8,6 @@ use crate::extraction::python::common::{ArgumentExtractor, ParameterFilter};
 use crate::extraction::{Parameter, ParameterValue, SdkMethodCall, SdkMethodCallMetadata};
 use crate::ServiceModelIndex;
 use ast_grep_language::Python;
-use convert_case::{Case, Casing};
 
 /// Information about a discovered get_waiter call
 #[derive(Debug, Clone)]
@@ -23,10 +22,21 @@ pub(crate) struct WaiterInfo {
     pub get_waiter_line: usize,
 }
 
+// TODO: This should be refactored at a higher level, so this type can be removed.
+// See https://github.com/awslabs/iam-policy-autopilot/issues/88.
 enum CallInfo<'a> {
     None(&'a WaiterInfo),
     Simple(&'a WaiterInfo, &'a WaitCallInfo),
     Chained(&'a ChainedWaiterCallInfo),
+}
+
+impl<'a> CallInfo<'a> {
+    fn waiter_name(&self) -> &'a str {
+        match self {
+            Self::None(waiter_info) | Self::Simple(waiter_info, ..) => &waiter_info.waiter_name,
+            Self::Chained(waiter_call_info) => &waiter_call_info.waiter_name,
+        }
+    }
 }
 
 /// Information about a wait method call
@@ -327,22 +337,15 @@ impl<'a> WaitersExtractor<'a> {
         wait_call: CallInfo,
         receiver: Option<String>,
     ) -> Vec<SdkMethodCall> {
-        let waiter_name = match wait_call {
-            CallInfo::None(waiter_info) => &waiter_info.waiter_name,
-            CallInfo::Simple(waiter_info, _) => &waiter_info.waiter_name,
-            CallInfo::Chained(waiter_call) => &waiter_call.waiter_name,
-        };
-
-        // Convert Python snake_case waiter name to PascalCase for lookup
-        let waiter_name_pascal = waiter_name.to_case(Case::Pascal);
         let mut synthetic_calls = Vec::new();
 
         // Get the operation for this service+waiter combination from service definition
-        if let Some(service_defs) = self.service_index.waiter_lookup.get(&waiter_name_pascal) {
+        if let Some(service_defs) = self
+            .service_index
+            .waiter_lookup
+            .get(wait_call.waiter_name())
+        {
             for service_method in service_defs {
-                // Convert operation name to Python snake_case for method name
-                let operation_snake = service_method.operation_name.to_case(Case::Snake);
-
                 let (parameters, start_position, end_position) = match wait_call {
                     CallInfo::Simple(_, wait_call) => {
                         // Filter out WaiterConfig from arguments - it's waiter-specific, not operation-specific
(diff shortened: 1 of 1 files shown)


iam-policy-autopilot-policy-generation/src/extraction/sdk_model.rs
```diff
@@ -360,7 +347,7 @@ impl ServiceDiscovery {
         // Collect results from parallel tasks
         while let Some(result) = join_set.join_next().await {
             match result {
-                Ok(Ok((service_info, mut service_model, waiters))) => {
+                Ok(Ok((service_info, service_model, waiters))) => {
```
adpaco, line 350: Nice

iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
```diff
@@ -317,114 +322,93 @@ impl<'a> WaitersExtractor<'a> {
...
-            .cloned()
-            .unwrap_or_default();
-
-        if candidate_services.is_empty() {
-            return Vec::new();
-        }
+        let waiter_name = match wait_call {
+            CallInfo::None(waiter_info) => &waiter_info.waiter_name,
+            CallInfo::Simple(waiter_info, _) => &waiter_info.waiter_name,
+            CallInfo::Chained(waiter_call) => &waiter_call.waiter_name,
```
adpaco: `waiter_name` should probably be a field on a high-level waiter extractor struct, and the variants here should be part of a `CallKind` enum. But we don't need to refactor that here.

Can we have a simple `impl` for `CallInfo`with a `waiter_name` method so we can do `let waiter_name = wait_call.waiter_name();`?

iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
```diff
@@ -317,114 +322,93 @@ impl<'a> WaitersExtractor<'a> {
...
+                let (parameters, start_position, end_position) = match wait_call {
+                    CallInfo::Simple(_, wait_call) => {
+                        // Filter out WaiterConfig from arguments - it's waiter-specific, not operation-specific
+                        (
+                            ParameterFilter::filter_waiter_parameters(wait_call.arguments.clone()),
+                            wait_call.start_position,
+                            wait_call.end_position,
+                        )
+                    }
+                    CallInfo::Chained(wait_call) => {
```
adpaco: Isn't this the same code? Can't we have a single arm for these two variants?
```rust
                    CallInfo::Simple(_, wait_call) | CallInfo::Chained(wait_call) => {
                        // Filter out WaiterConfig from arguments - it's waiter-specific, not operation-specific
                        (
                            ParameterFilter::filter_waiter_parameters(wait_call.arguments.clone()),
                            wait_call.start_position,
                            wait_call.end_position,
                        )
                    }

File: iam-policy-autopilot-policy-generation/src/extraction/typescript/extractor.rs [hunk 4]
```diff
@@ -416,4 +424,72 @@ const command = new GetObjectCommand({ Bucket: 'test', Key: 'test.txt' });
...
+        // Code that imports Neptune client and uses DBInstanceAvailable waiter
+        let typescript_code = r#"
+    import { NeptuneClient, waitUntilDBInstanceAvailable } from '@aws-sdk/client-neptune';
+
+    const client = new NeptuneClient({ region: 'us-east-1' });
+
+    async function waitForInstance() {
+        await waitUntilDBInstanceAvailable(
+            { client, maxWaitTime: 300 },
+            { DBInstanceIdentifier: 'my-neptune-instance' }
```
adpaco, line 444: Is this a real example? I couldn't find this API earlier.

iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
weibenz1: unrelated, why is it important to support waiters? Does it actually affect the permissions generated?

iam-policy-autopilot-policy-generation/src/extraction/typescript/extractor.rs
mschlaipfer: https://github.com/boto/botocore/blob/7c770fb210913ff9fa9b35470161c3278786085b/botocore/data/neptune/2014-10-31/waiters-2.json#L4

iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
mschlaipfer: @weibenz1 Yes, we generate the permissions needed for the waiter (e.g., [DBInstanceAvailable](https://boto3.amazonaws.com/v1/documentation/api/1.26.93/reference/services/rds/waiter/DBInstanceAvailable.html)) using the model [here](https://github.com/boto/botocore/blob/7c770fb210913ff9fa9b35470161c3278786085b/botocore/data/neptune/2014-10-31/waiters-2.json#L4). Not every operation has a waiter, and we also need to map the waiter name to the operations being used by the waiter to generate the right permissions.

iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
mschlaipfer: The types of `wait_call` are different, I'll rename to make it clear. We need to generalize these structs in a separate PR to get rid of a lot of duplication. I also ran into this duplication in https://github.com/awslabs/iam-policy-autopilot/pull/84.

iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
mschlaipfer: Introduced `waiter_name()`, resolving

iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
```diff
@@ -317,114 +332,88 @@ impl<'a> WaitersExtractor<'a> {
...
-    fn create_matched_synthetic_calls(
+    fn create_synthetic_calls_internal(
         &self,
-        wait_call: &WaitCallInfo,
-        waiter_info: &WaiterInfo,
+        wait_call: CallInfo,
+        receiver: Option<String>,
     ) -> Vec<SdkMethodCall> {
-        // Convert Python snake_case waiter name to PascalCase for lookup
-        let waiter_name_pascal = waiter_info.waiter_name.to_case(Case::Pascal);
```
mschlaipfer, line 326: This conversion here was wrong, fixed it as part of this PR. See https://github.com/awslabs/iam-policy-autopilot/issues/89

PR #81, opened 2025-12-15
docs: highlight service hints in README and help text

adpaco-aws:
*Description of changes:* Updated the README to recommend using service hints in the *Best Practices and .Considerations* section. Updated the CLI help text to use clearer language and show tips/usage of the `--service-hints` option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+## [0.1.2] - 2025-12-15
+
+## Fixed
+
+- Use SDK info to find the operation from a method name. Fixes a bug where `modify_db_cluster` (and similar names) was renamed incorrectly to `ModifyDbCluster` instead of `ModifyDBCluster`. (#70)
+- Reduce false positive findings by fixing Go SDK parameter extraction. It now uses required arguments correctly to disambiguate possible services. (#50)
+
+## Added
+
+- Added installation script for MacOS and Linux. (#44)
+
+## Changed
+
+- We now add the policy ID `IamPolicyAutopilot` in the access denied workflow.  (#48)
+- Updated Cargo.toml description. (#46)
+
 ## [0.1.1] - 2025-11-26
 
 ### 🚀 Features
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ exclude = ["iam-policy-autopilot-lints"]
 resolver = "2"
 
 [workspace.package]
-version = "0.1.1"
+version = "0.1.2"
 edition = "2021"
 license = "Apache-2.0"
 repository = "https://github.com/awslabs/iam-policy-autopilot"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "maturin"
 # Standard Python package metadata
 [project]
 name = "iam-policy-autopilot" # The name it will have on PyPI
-version = "0.1.1"
+version = "0.1.2"
 description = "An open source Model Context Protocol (MCP) server and command-line tool that helps your AI coding assistants quickly create baseline IAM policies that you can refine as your application evolves, so you can build faster. IAM Policy Autopilot analyzes your application code locally to generate identity-based policies for application roles, enabling faster IAM policy creation and reducing access troubleshooting time. IAM Policy Autopilot supports applications built in Python, Go, and TypeScript."
 readme = "README.md"
 requires-python = ">=3.8"


mschlaipfer (approved):

README.md
```diff
@@ -245,6 +261,7 @@ iam-policy-autopilot generate-policies \
 Options:
 - `--region <REGION>` - AWS region for resource ARNs
 - `--account <ACCOUNT>` - AWS account ID for resource ARNs
+- `--service-hints <SERVICES>` - Filter results to specific AWS services (reduces unnecessary permissions)
```
Dianayin422: I don't think this description makes senses to me. How about using what you wrote in main.rs? "limit analysis to only the services your application actually uses if you know of them. This helps reduce unnecessary permissions."

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -227,7 +228,10 @@ This flag has no effect on the generate-policies subcommand."
     #[command(long_about = "\
 Generates complete IAM policy documents from source files. By default, all \
 policies are merged into a single optimized policy document. \
-Optionally takes AWS context (region and account) for accurate ARN generation.")]
+Optionally takes AWS context (region and account) for accurate ARN generation.\n\n\
+TIP: Use --service-hints to specify only the AWS services your application uses. \
```
Dianayin422: How about "TIP: use --service-hints to specify the particular AWS services that your application uses if you know of them"?

README.md
weibenz1: yeah, maybe something like `Reduces unnecessary permissions by explicitly declaring which services your code interacts with`?

adpaco: Sounds good, we're using that description now (with "if you know them" which sounds more natural).

iam-policy-autopilot-cli/src/main.rs
adpaco: Sure, done.

README.md
adpaco: Just to be clear, we're using @Dianayin422 's proposal... (@weibenz1 's comment showed up just now for me). I believe @Dianayin422 's description is still more accurate, "services your code interacts with" may be too ambiguous in this case.

README.md
```diff
@@ -40,9 +40,25 @@ IAM Policy Autopilot stays up to date with the latest AWS services and features
 
 ## Best Practices and Considerations
 
+### Use service hints for accurate policies
```
Dianayin422, line 55: Can we move this to the be the last point in the "Best Practices and Considerations" section? This sounds more granular compared with other points.

README.md
```diff
@@ -40,9 +40,25 @@ IAM Policy Autopilot stays up to date with the latest AWS services and features
...
+
+IAM Policy Autopilot's static analysis may include permissions for AWS services your application doesn't use. This happens when method names in your code match AWS SDK calls from multiple services. For example, a method called `listAccounts()` might generate permissions for both IAM and Amazon Chime services.
```
Dianayin422: Here, can we link to both AWS Documentations page? I mean, IAM's listAccounts() and Amazon Chime's listAccounts(). So that people can easily find what you are referring to in this example. Are these 2 links correct:
https://docs.aws.amazon.com/organizations/latest/APIReference/API_ListAccounts.html 
https://docs.aws.amazon.com/chime/latest/APIReference/API_ListAccounts.html

README.md
```diff
@@ -40,9 +40,25 @@ IAM Policy Autopilot stays up to date with the latest AWS services and features
...
+
+**Recommended approach**: Use the `--service-hints` option to specify only the AWS services your application actually uses. This filters which SDK calls are analyzed, but the final policy may still include actions from other services if they're required by the operations you perform:
```
Dianayin422: Change to "This helps IAM Policy Autopilot scope down which SDK calls to analyze, but the final policy may still include actions from other services if they're required by the operations you perform:"

Dianayin422 (commented):
Left a few more comments.

README.md
mschlaipfer: Another thing: this suggestion applies to the CLI invocation. We might want to point out that the MCP server is already "configured" to use service hints (we expect the agent to populate it).

adpaco: @Dianayin422 updated.

@mschlaipfer added a note.

README.md
adpaco: Yes, also update to say "AWS Organizations" instead of "IAM" (technically more accurate).

README.md
adpaco: Done.

Dianayin422 (approved):

PR #84, opened 2025-12-17
feat: add `--explain`

mschlaipfer:
*Issue #, if available:*

*Description of changes:* This PR adds an experimental explanation feature to the CLI (`--explain`) which, for every action in the generated policy, provides a reason for why it has been added, including the expression and code location where IPA found an AWS operation call that led to its inclusion. It also provides information about [FAS expansion](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_forward_access_sessions.html), if it occurred.

We believe that this feature, together with `--service-hints`, will help creating policies with fewer unwanted permissions.

Example:
```json
"Explanations": {
    "ExplanationForAction": {
      "dynamodb:GetItem": [
        {
          "Operations": [
            {
              "Service": "dynamodb",
              "Name": "GetItem",
              "Source": {
                "Expr": "dynamodb.get_item(\n        TableName='my-table',\n        Key={'id': {'S': '123'}}\n    )",
                "Location": "iam-policy-autopilot-cli/tests/resources/test_example.py:19.5-22.6"
              }
            }
          ]
        }
      ],
      "kms:Decrypt": [
        {
          "Operations": [
            {
              "Service": "s3",
              "Name": "GetObject",
              "Source": {
                "Expr": "s3.get_object(Bucket='my-bucket', Key='my-file.txt')",
                "Location": "iam-policy-autopilot-cli/tests/resources/test_example.py:7.16-7.68"
              }
            },
            {
              "Service": "kms",
              "Name": "Decrypt",
              "Source": "FAS"
            }
          ]
        }
       ]
    },
    "Documentation": {
      "FAS": {
        "Plain": "The explanation contains an operation added due to Forward Access Sessions.",
        "URL": "https://docs.aws.amazon.com/IAM/latest/UserGuide/access_forward_access_sessions.html"
      }
    }
 }
```

This PR adds file and raw expression information and expands location information in multiple places throughout the extraction phase so this data is available in the `SdkMethodCall`. There is currently a lot of code duplication, which I have not cleaned up as part of this pull request. See https://github.com/awslabs/iam-policy-autopilot/issues/88.

During enrichment it adds the FAS expansion information. The location format follows the one [here](https://www.gnu.org/prep/standards/html_node/Errors.html). It is clickable, so it navigates to the location in VS Code (I have not tested with other editors).

This PR removes the (hidden) `--show-action-mappings` command which had a similar purpose, but did not support location, expression, or FAS information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## [0.1.3]
+## [Unreleased]
 
 ## Added
 
--- a/iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
@@ -211,47 +211,32 @@ pub struct Explanations {
     /// Explanation for inclusion of an action
     pub explanation_for_action: BTreeMap<String, Explanation>,
     /// Documentation of concepts used in the explanation for an action
-    pub documentation: BTreeMap<&'static str, Documentation>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub documentation: Vec<&'static str>,
 }
 
 impl Explanations {
-    const FAS_PLAIN: &str =
-        "The explanation contains an operation added due to Forward Access Sessions.";
-    const FAS_URL: &str =
-        "https://docs.aws.amazon.com/IAM/latest/UserGuide/access_forward_access_sessions.html";
+    const FAS: &str =
+        "The explanation contains an operation added due to Forward Access Sessions (FAS). See https://docs.aws.amazon.com/IAM/latest/UserGuide/access_forward_access_sessions.html.";
 
     pub(crate) fn new(explanations: BTreeMap<String, Explanation>) -> Self {
-        let mut documentation: Vec<(&'static str, Documentation)> = vec![];
+        let mut documentation: Vec<&'static str> = vec![];
         for explanation in explanations.values() {
             for reason in &explanation.reasons {
                 for op in &reason.operations {
                     match op.source {
                         OperationSource::Extracted(_) | OperationSource::Provided => (),
-                        OperationSource::Fas(_) => documentation.push((
-                            "FAS",
-                            Documentation {
-                                plain: Self::FAS_PLAIN,
-                                url: Self::FAS_URL,
-                            },
-                        )),
+                        OperationSource::Fas(_) => documentation.push(Self::FAS),
                     }
                 }
             }
         }
+        documentation.dedup();
         Self {
             explanation_for_action: explanations,
-            documentation: BTreeMap::from_iter(documentation),
+            documentation,
         }
     }
-}
-
-/// Documentation of concepts appearing in explanations
-#[derive(Debug, Clone, Serialize)]
-#[serde(rename_all = "PascalCase")]
-pub struct Documentation {
-    plain: &'static str,
-    #[serde(rename = "URL")]
-    url: &'static str,
 }
 
 /// Represents an explanation for why an action was added to a policy
--- a/iam-policy-autopilot-policy-generation/src/extraction/javascript/shared.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/javascript/shared.rs
@@ -196,7 +196,7 @@ impl ExtractionUtils {
                             // Extract operation name by removing "Command" suffix
                             if let Some(operation_name) = command_name.strip_suffix("Command") {
                                 // Keep PascalCase operation name to match service index
-                                // e.g., "CreateBucket" stays "CreateBucket"
+                                // e.g., "PutItem" from "PutItemCommand"
                                 let method_call = SdkMethodCall {
                                     name: operation_name.to_string(),
                                     possible_services: vec![service.clone()],
--- a/iam-policy-autopilot-policy-generation/src/extraction/mod.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/mod.rs
@@ -344,68 +344,10 @@ mod tests {
     }
 
     #[test]
-    fn test_parsed_method_creation() {
-        let method = SdkMethodCall {
-            name: "test_method".to_string(),
-            possible_services: Vec::new(),
-            metadata: Some(SdkMethodCallMetadata {
-                parameters: vec![Parameter::Keyword {
-                    name: "param1".to_string(),
-                    value: ParameterValue::Resolved("test_value".to_string()),
-                    position: 0,
-                    type_annotation: Some("str".to_string()),
-                }],
-                return_type: Some("bool".to_string()),
-                expr: "test_method".to_string(),
-                location: Location::new(PathBuf::new(), (10, 1), (10, 25)),
-                receiver: None,
-            }),
-        };
-
-        assert_eq!(method.name, "test_method");
-        assert_eq!(method.metadata.as_ref().unwrap().parameters.len(), 1);
-        assert_eq!(
-            method.metadata.as_ref().unwrap().location.start_position,
-            (10, 1)
-        );
-        assert_eq!(
-            method.metadata.as_ref().unwrap().location.end_position,
-            (10, 25)
-        );
-    }
-
-    #[test]
-    fn test_parsed_method_with_sdk_context() {
-        let method = SdkMethodCall {
-            name: "get_object".to_string(),
-            possible_services: vec!["s3".to_string()],
-            metadata: Some(SdkMethodCallMetadata {
-                parameters: vec![Parameter::Keyword {
-                    name: "Bucket".to_string(),
-                    value: ParameterValue::Resolved("my-bucket".to_string()),
-                    position: 0,
-                    type_annotation: Some("str".to_string()),
-                }],
-                return_type: Some("Dict[str, Any]".to_string()),
-                expr: "get_object".to_string(),
-                location: Location::new(PathBuf::new(), (15, 5), (15, 45)),
-                receiver: Some("s3_client".to_string()),
-            }),
-        };
-
-        assert_eq!(method.name, "get_object");
-        assert_eq!(
-            method.metadata.as_ref().unwrap().receiver,
-            Some("s3_client".to_string())
-        );
-        assert_eq!(method.possible_services, vec!["s3".to_string()]);
-        if let Parameter::Keyword {
-            value, position, ..
-        } = &method.metadata.as_ref().unwrap().parameters[0]
-        {
-            assert_eq!(value.as_string(), "my-bucket");
-            assert_eq!(*position, 0);
-        }
+    fn test_location_construction() {
+        let location = Location::new(PathBuf::new(), (10, 1), (10, 25));
+        assert_eq!(location.start_position, (10, 1));
+        assert_eq!(location.end_position, (10, 25));
     }
 
     #[test]


iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
```diff
@@ -85,43 +89,32 @@ pub async fn generate_policies(
...
         "Policy generation completed in {:?}, generated {} policies",
         total_duration,
-        policies.len()
+        result.policies.len()
     );
 
-    let mut final_policies = policies;
+    let mut final_policies = result.policies;
+    let explanations = if config.generate_explanations {
+        result.explanations
```
weibenz1, line 105: Does this mean we always execute the logic for producing explanations regardless user input?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +80,95 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
...
+        /// Detected service containing this operation
+        service: String,
+        /// Extracted expr
+        expr: String,
+        /// Location in source file
+        location: Location,
+    },
+    /// Operation provided (no metadata available)
+    #[serde(rename_all = "PascalCase")]
+    Provided {
```
weibenz1: what's provided? like service hints?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -8,14 +8,40 @@ use convert_case::{Case, Casing};
...
 use crate::enrichment::operation_fas_map::{FasOperation, OperationFasMap, OperationFasMaps};
 use crate::enrichment::service_reference::ServiceReference;
 use crate::enrichment::{Condition, ServiceReferenceLoader};
 use crate::errors::{ExtractorError, Result};
 use crate::service_configuration::ServiceConfiguration;
 use crate::{SdkMethodCall, SdkType};
 
+/// Represents an operation with its provenance chain (how we reached this operation via FAS expansion)
+#[derive(Debug, Clone)]
+struct OperationWithFasExpansion {
```
weibenz1: how is it different from FasInfo?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -95,37 +121,58 @@ impl ResourceMatcher {
...
+    ///
+    /// Returns operations with their FAS provenance chains showing how each operation was reached.
     fn expand_fas_operations_to_fixed_point(
         &self,
         initial: FasOperation,
-    ) -> Result<Vec<FasOperation>> {
-        let mut operations = HashSet::<FasOperation>::new();
-        operations.insert(initial);
+    ) -> Result<Vec<OperationWithFasExpansion>> {
+        let mut operations = HashSet::<OperationWithFasExpansion>::new();
```
weibenz1: nit: processed_operations?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -95,37 +121,58 @@ impl ResourceMatcher {
...
+
+                                let new_op_with_prov = OperationWithFasExpansion {
+                                    operation: additional_op.clone(),
+                                    fas_chain: new_chain,
+                                };
+
                                 // Only add if we haven't seen this operation before
-                                if !operations.contains(additional_op) {
-                                    newly_discovered.insert(additional_op.clone());
+                                if !operations.contains(&new_op_with_prov) {
```
weibenz1: shall we move this condition up to wrap this entire code block?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -224,9 +271,11 @@ impl ResourceMatcher {
         // Use fixed-point algorithm to safely expand FAS operations until no new operations are found
         let operations = self.expand_fas_operations_to_fixed_point(initial)?;
 
+        let mut explanations = vec![];
         let mut enriched_actions = vec![];
-        for operation in operations {
-            let service_name = operation.service(&self.service_cfg);
+
+        for op_with_fas_chain in operations {
```
weibenz1: do we really need to name it op_with_fas_chain?

every operation may optionally have fas chain right? so it's really just a normal operation.

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -278,22 +335,86 @@ impl ResourceMatcher {
                                 );
 
                                 enriched_actions.push(enriched_action);
+
+                                // Include FAS chain only if chain is non-empty
+                                let fas_info = if op_with_fas_chain.fas_chain.is_empty() {
+                                    log::debug!(
+                                        "  Action '{}': Excluding FAS chain (initial operation)",
```
weibenz1: what does `Excluding` mean here?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -278,22 +335,86 @@ impl ResourceMatcher {
...
+                                        action.name
+                                    );
+                                    None
+                                } else {
+                                    // Build full chain: fas_chain + current operation
+                                    let mut chain = op_with_fas_chain.fas_chain.clone();
```
weibenz1: should this be a method `.get_full_fas_chain()` implemented in the struct instead?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -278,22 +335,86 @@ impl ResourceMatcher {
...
+                                    log::debug!(
+                                        "  Action '{}': Including FAS chain: {:?}",
+                                        action.name,
+                                        chain
+                                    );
+                                    Some(FasInfo::new(chain))
+                                };
+
+                                // Create explanation for this action
+                                let explanation = Explanation {
```
weibenz1, line 266: I'd pass in a parameter so we skip unnecessary computation

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -278,22 +335,86 @@ impl ResourceMatcher {
...
                         // Fallback: operation action map does not exist, create basic action
                         if let Some(a) =
                             self.create_fallback_action(&parsed_call.name, &service_reference)?
                         {
-                            enriched_actions.push(a)
+                            let action_name = a.name.clone();
+                            enriched_actions.push(a);
+
+                            // Create explanation for fallback action
+                            let explanation = Explanation {
```
weibenz1: feels like Explanation should really be an optional field in the generated action

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: It's almost the same, just that the expansion in `FasInfo` also contains the operation, and the explanation URL. I considered implementing a custom serializer for `OperationWithFasExpansion`, but find that having `FasInfo` explicitly is cleaner.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
mschlaipfer: In case we add an API which bypasses the extraction phase, we could consider a version that just takes a list of operation with service names, not providing any metadata which we use for creating the explanation.

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: We need the operation with FAS chain first, to check `contains`, so we are computing the chain first. We use a custom `hash` implementation, that just hashes the `operation`, though. This is a bit ugly, I'll see if I can come up with something better.

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: Good callout, thanks

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: Computing this is cheap, let's optimize later.

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
mschlaipfer: Yes, I think that's okay for now. We can look into optimization later.

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: Didn't make sense, removed

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: Moved to `OperationWithFasExpansion::to_fas_info`

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: I pushed a new commit, but I didn't change this. Open for suggestions.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +80,108 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
...
+    /// The chain of operations in the FAS expansion
+    pub expansion: Vec<String>,
+}
+
+impl FasInfo {
+    /// Create a new FasInfo with the standard AWS documentation URL
+    #[must_use]
+    pub fn new(expansion: Vec<String>) -> Self {
+        Self {
+            explanation: FAS_URL.to_string(),
```
adpaco: Can we avoid creating these strings by having a static string here? Or do you envision `explanation` being something else down the line?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +80,108 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
...
+        /// Extracted expr
+        expr: String,
+        /// Location in source file
+        location: Location,
+    },
+    /// Operation provided (no metadata available)
+    #[serde(rename_all = "PascalCase")]
+    Provided {
+        /// Provided name
+        name: String,
```
adpaco: `OperationView` could just be a struct with `name`, `service` and optional metadata.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +80,108 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
...
+        }
+    }
+}
+
+/// Represents an explanation for why an action was added to a policy
+#[derive(Debug, Clone, Serialize, PartialEq, Eq, Hash, JsonSchema, Default)]
+#[serde(rename_all = "PascalCase")]
+pub struct Explanation {
+    /// The reasons this action was added (can have multiple reasons for the same action)
+    pub reasons: Vec<Reason>,
```
adpaco, line 248: Could we end up with the same reason multiple times?

iam-policy-autopilot-policy-generation/src/extraction/javascript/shared.rs
```diff
@@ -48,6 +49,50 @@ fn load_libraries_mapping() -> Option<JsV3LibrariesMapping> {
     serde_json::from_str(content).ok()
 }
 
+/// Result of finding a command/function instantiation with its arguments
+#[derive(Debug, Clone)]
+pub(crate) struct CommandUsage<'a> {
```
adpaco, line 55: I'm confused: Why do we need this? Why didn't we need it until now?

mschlaipfer: I believe I've introduced it due to the complex return type of `find_command_instantiation_with_args`

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
mschlaipfer: Not after applying `merge`, but I think making this a set makes sense.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +80,108 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
...
+    #[serde(rename_all = "PascalCase")]
+    Provided {
+        /// Provided name
+        name: String,
+        /// Provided service
+        service: String,
+    },
+}
+
+impl OperationView {
```
weibenz1: maybe I misunderstood something, but this implementation still makes me feel we are doing a backward computation.

Instead of attaching reasons as part of the process for generating policy/action, we are computing/attaching the reasons after the action is generated, which seems odd to me

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +80,108 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
...
+    pub operation: OperationView,
+    /// FAS (Forward Access Sessions) expansion information if this action came from FAS expansion
+    #[serde(rename = "FAS", skip_serializing_if = "Option::is_none")]
+    pub fas: Option<FasInfo>,
+}
+
+#[derive(Debug, Clone, Serialize, PartialEq, Eq, Hash, JsonSchema)]
+#[serde(rename_all = "PascalCase")]
+#[serde(untagged)]
+pub enum OperationView {
```
weibenz1: nit: OperationSource?

If we name it as OperationSource, can we embed FasInfo inside this enum instead?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
weibenz1: I'm guessing @mschlaipfer wanted to achieve oneOf. 

What about a struct with `name` and `service` along with an enum OperationSource: [Extracted, Provided, FasExpansion]?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -96,16 +246,35 @@ impl Action {
...
     /// * `resources` - List of enriched resources
     /// * `conditions` - List of conditions
+    /// * `explanation` - Explanation why the action has been added
     #[must_use]
-    pub(crate) fn new(name: String, resources: Vec<Resource>, conditions: Vec<Condition>) -> Self {
+    pub(crate) fn new(
+        name: String,
+        resources: Vec<Resource>,
+        conditions: Vec<Condition>,
+        explanation: Explanation,
```
weibenz1, line 333: thanks!

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -96,16 +246,35 @@ impl Action {
...
     }
 }
 
+/// Represents a resource enriched with ARN pattern and metadata
+///
+/// This structure combines OperationAction resource data with Service Reference ARN patterns and additional
+/// metadata to provide complete resource information for IAM policies.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub(crate) struct Resource {
+    /// The resource type name (e.g., "bucket", "object", "*")
```
weibenz1, line 350: this confuses me a little bit - is it field modelling resource type or the name?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -8,14 +8,55 @@ use convert_case::{Case, Casing};
...
 use crate::enrichment::service_reference::ServiceReference;
 use crate::enrichment::{Condition, ServiceReferenceLoader};
 use crate::errors::{ExtractorError, Result};
 use crate::service_configuration::ServiceConfiguration;
 use crate::{SdkMethodCall, SdkType};
 
+/// Represents an operation with its provenance chain (how we reached this operation via FAS expansion)
+#[derive(Debug, Clone)]
+struct OperationWithFasExpansion {
+    operation: FasOperation,
```
weibenz1: If we had `GeneratedAction`, we don't need these different structs for modeling this variant of operation.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -67,26 +227,16 @@ pub(crate) trait Context {
...
-pub(crate) struct Resource {
-    /// The resource type name (e.g., "bucket", "object", "*")
-    pub(crate) name: String,
-    /// ARN patterns from Service Reference data, if available
-    pub(crate) arn_patterns: Option<Vec<String>>,
+    /// Optional explanation why this action has been added
+    pub(crate) explanation: Explanation,
 }
 
 impl Action {
```
weibenz1, line 320: Not for this PR - I'm starting to feel that we really should have just a struct for all generated operation/action like `GeneratedAction`/`GeneratedOperation`, and have fields like `source` annotating additional information and use this single struct only to model all operations. 

Probably will make our code more maintainable and clearer. 

What do you think?

mschlaipfer: I'm implementing this refactoring

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
mschlaipfer: I am about to change this in the next revision

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
mschlaipfer: The `name` is not used, will remove in separate PR. https://github.com/awslabs/iam-policy-autopilot/issues/99

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: This is resolved given the large refactoring of the new version

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
mschlaipfer: I considered making this a set, but I would like to maintain the order between runs, so would need a `BTreeSet` with an `Ord` implementation for `Reason`, which I'm not sure about. I will leave this as-is for now.

iam-policy-autopilot-policy-generation/src/service_configuration.rs
```diff
@@ -14,6 +14,8 @@ use std::{
 
 /// Operation rename configuration
 #[derive(Clone, Debug, Deserialize)]
+// TODO: remove
+#[allow(dead_code)]
```
adpaco, line 18: What's preventing us from removing it? Do we need to track this work using an issue?

iam-policy-autopilot-policy-generation/src/extraction/javascript/shared.rs
```diff
@@ -155,15 +196,15 @@ impl ExtractionUtils {
                             // Extract operation name by removing "Command" suffix
                             if let Some(operation_name) = command_name.strip_suffix("Command") {
                                 // Keep PascalCase operation name to match service index
-                                // e.g., "PutItem" from "PutItemCommand"
+                                // e.g., "CreateBucket" stays "CreateBucket"
```
adpaco: Why delete the other example?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +33,248 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
 pub(crate) use resource_matcher::ResourceMatcher;
 pub(crate) use service_reference::RemoteServiceReferenceLoader as ServiceReferenceLoader;
 
+/// Represents the reason why an action was added to a policy
+#[derive(derive_new::new, Debug, Clone, Serialize, PartialEq, Eq, Hash, JsonSchema)]
+#[serde(rename_all = "PascalCase")]
+pub struct Reason {
+    /// The original operation that was extracted
+    pub operations: Vec<Arc<Operation>>,
```
adpaco, line 41: I think the `Hash` impl might not work as expected because of how `Arc<T>` implements it?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +33,248 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
...
+}
+
+#[derive(Debug, Clone, Serialize, Eq, JsonSchema)]
```
adpaco, line 44: You need `PartialEq`

iam-policy-autopilot-policy-generation/src/service_configuration.rs
mschlaipfer: It's unrelated to the change, I've created https://github.com/awslabs/iam-policy-autopilot/issues/102.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
mschlaipfer: Custom implementation further down

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
mschlaipfer: Could you elaborate? I'll check if we even use `Hash`

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## [0.1.3]
```
weibenz1: this should happen when we create the new release, not as part of feature PR, right?

iam-policy-autopilot-cli/src/output.rs
```diff
@@ -1,8 +1,6 @@
 use anyhow::{Context, Result};
 use iam_policy_autopilot_access_denied::{DenialType, PlanResult};
-use iam_policy_autopilot_policy_generation::policy_generation::{
-    MethodActionMapping, PolicyWithMetadata,
-};
+use iam_policy_autopilot_policy_generation::api::model::GeneratePoliciesResult;
```
weibenz1, line 3: nice!

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -1,23 +1,35 @@
...
+    /// Generate explanations for why actions were added
+    pub generate_explanations: bool,
+}
+
+/// Result of policy generation including policies, action mappings, and explanations
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct GeneratePoliciesResult {
+    /// Generated IAM policies
+    pub policies: Vec<PolicyWithMetadata>,
```
weibenz1, line 29: Q: do you think if it makes sense to embed Explaination within PolicyWithMetadata struct?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +33,248 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
...
+                service,
+                name,
+                source: OperationSource::Extracted(metadata.clone()),
+                _private: (),
+            },
+        })
+    }
+}
+
+impl From<FasOperation> for Operation {
```
weibenz1, line 134: Q: are we keeping FasOperation for data modeling purpose? Or ask in a different way, can we refactor FasOperation with Operation completely?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +33,248 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
...
+            .to_string();
+        let name = if sdk == SdkType::Boto3 {
+            // Try to load service reference and look up the boto3 method mapping
+            service_reference_loader
+                .load(&service)
+                .await?
+                .and_then(|service_ref| {
+                    log::debug!("Looking up method {}", call.name);
+                    service_ref
+                        .boto3_method_to_operation
```
weibenz1, line 99: shall we implement this function direction within service reference loader?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +33,248 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
...
+        "The explanation contains an operation added due to Forward Access Sessions.";
+    const FAS_URL: &str =
+        "https://docs.aws.amazon.com/IAM/latest/UserGuide/access_forward_access_sessions.html";
+
+    pub(crate) fn new(explanations: BTreeMap<String, Explanation>) -> Self {
+        let mut documentation: Vec<(&'static str, Documentation)> = vec![];
+        for explanation in explanations.values() {
+            for reason in &explanation.reasons {
+                for op in &reason.operations {
+                    match op.source {
```
weibenz1, line 227: if we inject the explaination at the time of computing/creating operation and operation source, we'd not need to do this reverse-computation right?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -4,128 +4,62 @@
...
-/// combining parsed method calls with operation action maps and Service
-/// Definition Files to produce complete IAM metadata.
-#[derive(Debug, Clone)]
-pub(crate) struct ResourceMatcher {
-    service_cfg: Arc<ServiceConfiguration>,
-    fas_maps: OperationFasMaps,
-    sdk: SdkType,
+#[derive(Clone, Debug)]
+struct FasExpansion {
+    dependency_graph: HashMap<Arc<Operation>, Vec<Arc<Operation>>>,
```
weibenz1, line 20: Q: why does it have to be wrapped by ARC, for cloning purposes?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -154,14 +83,87 @@ impl ResourceMatcher {
...
+impl ResourceMatcher {
+    /// Enrich a parsed method call with OperationAction maps, FAS maps, and Service
+    /// Reference data
+    pub(crate) async fn enrich_method_call<'b>(
+        &self,
+        parsed_call: &'b SdkMethodCall,
+        service_reference_loader: &ServiceReferenceLoader,
+    ) -> Result<Vec<EnrichedSdkMethodCall<'b>>> {
+        if parsed_call.possible_services.is_empty() {
+            return Err(ExtractorError::enrichment_error(
```
weibenz1, line 147: Q: is this always considered to be an error?

iam-policy-autopilot-policy-generation/src/policy_generation/engine.rs
```diff
@@ -247,60 +251,57 @@ impl<'a> Engine<'a> {
...
-            };
+        Ok(GeneratePoliciesResult {
+            policies,
+            explanations: Some(explanations),
+        })
+    }
+}
 
-            mappings.push(method_mapping);
+fn extract_explanations(enriched_calls: &[EnrichedSdkMethodCall<'_>]) -> Explanations {
```
weibenz1, line 283: I still think having explanation embedded within GeneratedAction/Action will make everything cleaner, unless I'm missing something here?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -22,6 +33,248 @@ pub(crate) use operation_fas_map::load_operation_fas_map;
...
+                            log::debug!("got {:?}", op);
+                            op.split(':').nth(1).unwrap_or(op).to_string()
+                        })
+                })
+                // Fallback to PascalCase conversion if mapping not found
+                // This should not be reachable, but if for some reason we cannot use the SDF,
+                // we try converting to PascalCase, knowing that this is flawed in some cases:
+                // think `AddRoleToDBInstance` (actual name)
+                //   vs. `AddRoleToDbInstance` (converted name)
+                .unwrap_or_else(|| call.name.to_case(Case::Pascal))
```
adpaco, line 111: I have mixed feelings about this. I understand wanting to do a best-effort attempt at finding the call. Would this be one of the special cases for #89 ?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
adpaco: Nevermind, I didn't notice `Operation` also has a custom `Hash` implementation further down.

iam-policy-autopilot-policy-generation/src/extraction/mod.rs
```diff
@@ -359,8 +387,8 @@ mod tests {
                     type_annotation: Some("str".to_string()),
                 }],
                 return_type: Some("Dict[str, Any]".to_string()),
-                start_position: (15, 5),
-                end_position: (15, 45),
+                expr: "get_object".to_string(),
```
adpaco: My understanding was that `expr` would include the complete text of the SDK call, including sub-arguments and such, and be as long as the offset between the two positions (e.g., in this case we same line and columns 45-5 so it should be 40 characters). Is that correct? Or what am I missing?

CHANGELOG.md
mschlaipfer: ~~How/when do we populate these? It shouldn't be on the person performing the release to come up with these blurbs.~~

You probably just meant the version number, yes that can only be a placeholder at this point. Now checked https://keepachangelog.com, and it should just go into an `Unreleased` section. Didn't know!

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
mschlaipfer: Yes, my thought was to keep it for modeling the data cleanly, and then have a conversion into `Operation`. Do you have concerns with that?

iam-policy-autopilot-policy-generation/src/api/model.rs
mschlaipfer: I can see a point made for it. In particular, it might make it harder for the implementations for computing the policies and computing the explanations to diverge. If we move the explanation logic to `PolicyWithMetadata`, we need to also to make the `PolicyMerger` aware of the explanations.

I believe that presenting the explanations separately (and not per-policy) is the right thing to do, though. With that I'm in favor to keep the explanation computation as-is, as the logic is simple and local right now.

Thoughts?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: Yes, I wrapped it in an `Arc` for that reason, as the `Operation` also might have call metadata information, so I assume cloning it is expensive.

iam-policy-autopilot-policy-generation/src/policy_generation/engine.rs
mschlaipfer: Answered above.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
mschlaipfer: Could you create a separate issue? I'm okay with changing this, but it's just a code move in this PR.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
mschlaipfer: Seems like a good idea. As part of that, we should add an `sdk_type` to `SdkMethodCall`. Created https://github.com/awslabs/iam-policy-autopilot/issues/105

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
mschlaipfer: Answered elsewhere in more detail. If we compute the explanation as part of the policy, we'd need to thread it through the policy merger, and then merge each policy's explanation to generate a single explanation (and not one per policy).

The logic here in particular is scanning for any FAS operations in the explanation to generate documentation, if needed.

iam-policy-autopilot-policy-generation/src/extraction/mod.rs
mschlaipfer: You're right, this doesn't make sense. It's dummy data which is wrong. I think this test is pointless, I'll remove it.

mschlaipfer: Similarly, changed another test, to instead just test `Location::new`, as it was relying solely on standard `struct` construction syntax.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
weibenz1: I think that's fine, just curious to understand if there's any other use

iam-policy-autopilot-policy-generation/src/api/model.rs
weibenz1: I think it makes sense from the output format perspective; but the internals can still be done with Explanation embedded into PolicyWithMetadata.

But i'm totally okay with keeping it separate for now given our usage here

weibenz1 (approved):

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
adpaco: No need, we'll likely discuss this in #91 once it's updated.

PR #86, opened 2025-12-17
fix: update service reference user agent for integration tests

karanjitsingh:
*Description of changes:*

~Use a separate user agent when calling for service reference during integration tests.~
- ~Creates and env variable in `.cargo/cargo.toml` - this env var is picked up when running in the workspace through `cargo run` / `cargo test`~
- ~Updates pr checks~

Update: Change introduces a feature `integ-test` in `iam-policy-autopilot-policy-generation` for integration tests. All other crates containing integration tests use this feature in their `dev-dependencies`

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,28 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Please share the command, input, and output that can be used to reproduce the bug.
+
+Note: Please make sure to redact sensitive information from your policies, ARNs, and account IDs.
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Environment (please complete the following information):**
+ - OS: [e.g., macOS, Linux, Windows]
+ - IAM Policy Autopilot Version: [e.g., 0.1.0] (run `iam-policy-autopilot --version` or `uvx iam-policy-autopilot --version`)
+ - Installation Method: [e.g., uvx, pip, direct install]
+ - MCP Client (if applicable): [e.g., Kiro, Claude Desktop, N/A]
+
+**Additional context**
+Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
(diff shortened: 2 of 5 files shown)


.cargo/config.toml
```diff
@@ -0,0 +1,4 @@
+# Cargo configuration for the workspace
+
+[env]
+IAM_POLICY_AUTOPILOT_INTEGRATION_TEST = "1"
```
weibenz1: Q: doesn't this always set the environment value to be 1?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -385,8 +385,13 @@ impl RemoteServiceReferenceLoader {
...
 
     fn create_client() -> crate::errors::Result<Client> {
-        let user_agent = format!("{}/{}", IAM_POLICY_AUTOPILOT, env!("CARGO_PKG_VERSION"));
+        let user_agent_suffix = if std::env::var("IAM_POLICY_AUTOPILOT_INTEGRATION_TEST").is_ok() {
+                "-integration-test"
+        } else {
+            ""
+        };
 
+        let user_agent = format!("{}/{}{}", IAM_POLICY_AUTOPILOT, user_agent_suffix, env!("CARGO_PKG_VERSION"));
```
nwinans-aws: should the format string be `{}{}/{}` instead to preserve the `/version` ending

.github/workflows/pr-checks.yml
```diff
@@ -83,6 +83,8 @@ jobs:
 
       - name: Run tests
         if: matrix.target == 'x86_64-unknown-linux-gnu'
+        env:
```
weibenz1: we should also update other actions

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -385,8 +385,13 @@ impl RemoteServiceReferenceLoader {
     }
 
     fn create_client() -> crate::errors::Result<Client> {
-        let user_agent = format!("{}/{}", IAM_POLICY_AUTOPILOT, env!("CARGO_PKG_VERSION"));
+        let user_agent_suffix = if std::env::var("IAM_POLICY_AUTOPILOT_INTEGRATION_TEST").is_ok() {
```
weibenz1: I think you'd want to keep this logic but remove the config file, and only set it in github actions instead?

weibenz1:
I'd prefer to have a feature like `integration_test` or something in the enrichment crate, and integration test crate can just enable it
```
#[cfg(feature = "integ_test")]
pub fn override_user_agent() {
    ...
}
```

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
karanjitsingh: config file will also ensure locally run integration tests also run with the same user agent

.github/workflows/pr-checks.yml
weibenz1: no longer needed?

karanjitsingh: removed

weibenz1 (approved):

Issue #82, opened 2025-12-15
Expose boto3, botocore git commit + data hash information as part of `--version` output


hwei0:
During `build.rs` stage:
- Get the git commit hash and tag (if present) for each of `boto3` and `botocore` submodules
- Compute a hash over all the simplified files in `boto3-data-simplified` and `botocore-data-simplified`
- Write to a intermediate file, all this info for each of `boto3`, `botocore`

In CLI, upon a call to `--version`:
- Read the respective intermediate file, retrieve the commit hash, git tag, and data hash
- Output to command line the following, in addition to existing version information: `{boto3/botocore} version: commit={commit_hash}, tag={tag if tag is present else None}, data_hash={data_hash}`

adpaco-aws:
Instead of separate version commands for these, I'd recommend printing the information as part of `--version` (I believe the version command behavior can be overriden in `clap`).

weibenz1:
> tag={tag if tag is present else None}

I thought we'd always be monitoring tags rather than latest commit - in this case when would we not have it?

hwei0:
When I pulled and submodule updated a few hours ago, there werent any tags with that commit it seemed

PR #87, opened 2025-12-17
feat: add submodule version+data hash info to --version --debug output

hwei0:
*Issue #, if available:*
#82 

*Description of changes:*

added custom `--version` output to publish boto3, botocore git tag+commit+data hash (data hash is the hash of the simplified data files). **This information will only be exposed if the hidden `--debug` flag is set; current --version behaviour without the `--deubg` flag will remain unchanged.**

example output
```
./target/debug/iam-policy-autopilot --version --debug
0.1.2
boto3 version: commit_id=0ca53211df7d2bb2213436a6b19e8cf0b44dee7d, commit_tag=1.42.11, data_hash=8309474F524E2F3DE2F90ADA66F1C5C887358CC503846073663BA311783FF495
botocore version: commit_id=7c770fb210913ff9fa9b35470161c3278786085b, commit_tag=None, data_hash=615E77E2C8C1DD6FD471416D712F4FE5A8E66B45150DC5D4BDC192736A8A42F1
```

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/shared_submodule_model.rs
+++ b/iam-policy-autopilot-policy-generation/src/shared_submodule_model.rs
@@ -1,12 +1,14 @@
-
 /// Exposes git version and commit hash for boto3 and botocore
+/// The struct defined here is used in both build.rs and model.rs.
+/// To share this struct in both the library and the build step, we define it here,
+/// and use include!(...) to include it in both uses.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "PascalCase")]
 pub struct GitSubmoduleMetadata {
-    /// the commit of boto3/botocore, returned on calls to iam-policy-autopilot --version --debug
+    /// the commit of boto3/botocore, returned on calls to iam-policy-autopilot --version --verbose
     pub git_commit_hash: String,
-    /// the git tag of boto3/botocore, returned on calls to iam-policy-autopilot --version --debug
+    /// the git tag of boto3/botocore, returned on calls to iam-policy-autopilot --version --verbose
     pub git_tag: Option<String>,
-    /// the sha hash of boto3/botocore simplified models, returned on calls to iam-policy-autopilot --version --debug
+    /// the sha hash of boto3/botocore simplified models, returned on calls to iam-policy-autopilot --version --verbose
     pub data_hash: String,
 }


iam-policy-autopilot-cli/src/commands.rs
```diff
@@ -143,6 +143,29 @@ async fn fix_access_denied_with_service(
...
 }
 
+pub fn print_version_info() -> anyhow::Result<()> {
+    let boto3_version_metadata =
+        iam_policy_autopilot_policy_generation::api::get_boto3_version_info()?;
+    let botocore_version_metadata =
+        iam_policy_autopilot_policy_generation::api::get_botocore_version_info()?;
+    println!("{}", crate_version!());
+    println!(
+        "boto3 version: commit_id={}, commit_tag={}, data_hash={}",
```
weibenz1: Q: is this something we'd want to expose to customers?

iam-policy-autopilot-cli/src/commands.rs
```diff
@@ -143,6 +143,29 @@ async fn fix_access_denied_with_service(
     }
 }
 
+pub fn print_version_info() -> anyhow::Result<()> {
+    let boto3_version_metadata =
+        iam_policy_autopilot_policy_generation::api::get_boto3_version_info()?;
+    let botocore_version_metadata =
+        iam_policy_autopilot_policy_generation::api::get_botocore_version_info()?;
+    println!("{}", crate_version!());
```
weibenz1, line 147: I'd make this version more specific so user understands that this is for IPA version

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -123,6 +150,60 @@ fn main() {
...
+    let boto3_info_json =
+        serde_json::to_string(&boto3_info).expect("Failed to serialize boto3 version metadata");
+    fs::write(boto3_submodule_version_dir, boto3_info_json)
+        .expect("Failed to write boto3 version metadata");
+
+    let botocore_submodule_dir = Path::new("resources/config/sdks/botocore-data");
+    let botocore_repo =
+        Repository::open(botocore_submodule_dir).expect("Failed to open botocore repository");
+
+    let botocore_info = GitSubmoduleVersion {
```
weibenz1: shall we implement all the version computing logic as methods within GitSubmoudleVersion, so we just need to initialize the struct which then auto compute everything?

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -344,6 +425,47 @@ fn copy_dir_recursive(src: &Path, dst: &Path) -> Result<(), Box<dyn std::error::
     Ok(())
 }
 
+fn sha2sum_recursive(
```
weibenz1: I'd move this to be part of GitSubmoduleVersion as well

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -49,6 +51,20 @@ pub struct AwsContext {
     pub account: String,
 }
 
+/// Exposes git version and commit hash for boto3 and botocore
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct GitSubmoduleMetadata {
```
weibenz1: can we consolidate `GitSubmoduleMetadata` and `GitSubmoduleVersion`?

Cargo.toml
```diff
@@ -46,7 +46,7 @@ criterion = "0.5"
 proptest = "1.0"
```
weibenz1, line 47: Whatever final format we agree on, we'd also need to update the github actions that validate versions.

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -344,6 +425,47 @@ fn copy_dir_recursive(src: &Path, dst: &Path) -> Result<(), Box<dyn std::error::
...
+    cwd: &Path,
+    root: &Path,
+) -> Result<
+    GenericArray<u8, UInt<UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>, B0>>,
```
mschlaipfer: What's this type? Let's introduce a type alias.

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -344,6 +425,47 @@ fn copy_dir_recursive(src: &Path, dst: &Path) -> Result<(), Box<dyn std::error::
...
+    Box<dyn std::error::Error>,
+> {
+    let mut hash_table: BTreeMap<
+        RelativePathBuf,
+        GenericArray<u8, UInt<UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>, B0>>,
+    > = BTreeMap::new();
+    // let next_root = if (root.is_none()) {Some(cwd)} else {root};
```
mschlaipfer: No commented code

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -123,6 +150,60 @@ fn main() {
...
+    let boto3_repo =
+        Repository::open(&boto3_submodule_dir).expect("Failed to open boto3 repository");
+
+    let boto3_info = GitSubmoduleVersion {
+        git_commit_hash: get_repository_commit(&boto3_repo)
+            .expect("Failed to get boto3 repository commit"),
+        git_tag: get_repository_tag(&boto3_repo).expect("Failed to get boto3 repository tag"),
+        data_hash: format!(
+            "{:X}",
+            sha2sum_recursive(&boto3_dir, &boto3_dir)
```
mschlaipfer: Why does this take the same argument twice?

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -123,6 +150,60 @@ fn main() {
...
+    let botocore_repo =
+        Repository::open(botocore_submodule_dir).expect("Failed to open botocore repository");
+
+    let botocore_info = GitSubmoduleVersion {
+        git_commit_hash: get_repository_commit(&botocore_repo)
+            .expect("Failed to get botocore repository commit"),
+        git_tag: get_repository_tag(&botocore_repo).expect("Failed to get botocore repository tag"),
+        data_hash: format!(
+            "{:X}",
+            sha2sum_recursive(&simplified_dir, &simplified_dir)
```
mschlaipfer: Seems like this method takes the same argument twice in each invocation.

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -49,6 +51,20 @@ pub struct AwsContext {
...
+    /// the git commit hash
+    #[serde(rename = "gitCommit")]
```
mschlaipfer: We use `PascalCase` for JSON throughout the project, let's use `#[serde(rename_all="PascalCase")]` on the struct and then specialize for `GitCommit`

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -417,3 +539,29 @@ fn process_boto3_service_version(
...
+}
+
+fn get_repository_commit(repo: &Repository) -> Result<String, Box<dyn std::error::Error>> {
+    Ok(repo
+        .revparse_single("HEAD")?
+        .into_commit()
+        .expect("Failed to get HEAD commit hash")
+        .id()
+        .to_string())
+}
```
mschlaipfer, line 539: Would it make sense to structure this file by introducing a separate `mod` for the version info computation, and one for the data simplification/copying?

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -1,18 +1,34 @@
+use git2::Commit;
+use git2::Describe;
+use git2::DescribeFormatOptions;
+use git2::DescribeOptions;
+use git2::Reference;
+use git2::Repository;
```
mschlaipfer: I'd prefer grouping here, at least on the entity after the last `::` like `use git2::{Commit, Describe, ...}`. Maybe `cargo fmt` doesn't do that? At least we should be consistent.

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -49,6 +65,17 @@ struct ShapeReference {
     shape: String,
 }
 
+// Git version and commit hash for boto3 and botocore
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct GitSubmoduleVersion {
+    #[serde(rename = "gitCommit")]
```
mschlaipfer: Same comment as on the other similar struct regarding `PascalCase`. I assume we cannot easily share these structs as this is in `build.rs`?

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -49,6 +65,17 @@ struct ShapeReference {
...
+    git_commit_hash: String,
+    #[serde(rename = "gitTag")]
+    git_tag: Option<String>,
+    #[serde(rename = "dataHash")]
+    data_hash: String,
```
mschlaipfer: Could you add a code comment about why we need commit hash, tag, and data hash?

iam-policy-autopilot-cli/src/commands.rs
mschlaipfer: We could have a separate command which we hide, that is easier to parse, it could output JSON.

Cargo.toml
```diff
@@ -64,7 +64,9 @@ serial_test = "3.0"
 atty = "0.2"
 chrono = { version = "0.4", features = ["serde"] }
 uuid = { version = "1.8", features = ["v4"] }
-sha2 = "0.10"
+sha2 = "0.10.9"
```
mschlaipfer: I think this is causing the OpenSSL build issue. We use `rustls-tls` in the `reqwest` dependency to work around a similar issue. I wonder if using use https://docs.rs/aws-lc-rs/1.15.2/aws_lc_rs/digest/fn.digest.html here would resolve the error?

mschlaipfer:
The PR title will need a `chore:` or `feat:` so it gets merged correctly (I don't think we have the PR title check, which should block merging this, yet).

iam-policy-autopilot-policy-generation/build.rs
weibenz1: it should do it, the PR checks failed on formatting check

iam-policy-autopilot-policy-generation/src/embedded_data.rs
```diff
@@ -31,6 +33,11 @@ struct BotocoreRaw;
 #[include = "*.json"]
 struct Boto3ResourcesRaw;
 
+#[derive(RustEmbed)]
+#[folder = "target/submodule-version-info"]
+#[include = "*.json"]
+struct GitSubmoduleVersionInfoRaw;
```
adpaco, line 38: You could avoid embedding this data and parsing it in JSON by using the `rustc-env` instruction in the build script: https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env - You'd only need to set the environment variables (e.g., BOTO3_SUBMODULE_TAG`) to whatever you compute in the script, then extract them here.

iam-policy-autopilot-cli/src/commands.rs
hwei0: ack, i'll make a new `--debug` hidden flag to do this. 

i'm against JSON for now (easier to parse line-by-line with `sed` and `| head -n X | tail -n 1`), but if we want JSON we can revisit later.

iam-policy-autopilot-cli/src/commands.rs
hwei0: from [this](https://github.com/awslabs/iam-policy-autopilot/pull/87#discussion_r2636090675), i'll make `--version` the same, but surface a different `--debug` hidden flag to show the full output (i.e. with dependency versions) if its set.

Cargo.toml
hwei0: from [this](https://github.com/awslabs/iam-policy-autopilot/pull/87#discussion_r2636090675), i'll make `--version` the same as before, so no need to update

iam-policy-autopilot-policy-generation/build.rs
hwei0: ack will change. for context, its the output type of sha256 :)

iam-policy-autopilot-policy-generation/build.rs
hwei0: the first arg is the directory that sha256 currently iterates on. For recursive steps into subdirectories, the second argument is needed to keep track of the root directory (which is used to compute relative paths that are inputted into the sha computation)

iam-policy-autopilot-policy-generation/build.rs
hwei0: that's intended, the second argument is used in recursive sub-calls after the initial call.

iam-policy-autopilot-policy-generation/src/api/model.rs
hwei0: hmm ok i'll do that, but the other structs in `build.rs` don't seem to have this.

iam-policy-autopilot-policy-generation/build.rs
hwei0: yeah makes a lot of sense, i'll change

update: more in favor of just moving all the verison info computation to be within GitSubmoduleVersion

iam-policy-autopilot-policy-generation/build.rs
hwei0: ack re: pascalCase. Yeah these structs don't seem to be easily sharable, for that reason.

iam-policy-autopilot-policy-generation/src/api/model.rs
mschlaipfer: okay, feel free to skip, if it's not needed here

Cargo.toml
hwei0: i'll try!

iam-policy-autopilot-policy-generation/src/embedded_data.rs
hwei0: i see, that's good to know for the future. I'm opposed to doing that here since there's six such tag/hash/commit values, which might justify putting them in an embed struct instead.

iam-policy-autopilot-policy-generation/build.rs
hwei0: (i did run `cargo fmt`...) but will fix

iam-policy-autopilot-policy-generation/src/api/model.rs
hwei0: don't think that's quite easy to do, since it seems build.rs can't import from the actual library files (as it runs during pre-compilation)

Cargo.toml
hwei0: seems like `aws-lc-rs` also [failed the build](https://github.com/awslabs/iam-policy-autopilot/actions/runs/20384413958) for the same reason :skull:

i [resolved the build failure](https://github.com/awslabs/iam-policy-autopilot/actions/runs/20384672870/job/58583031327) by adding the following to Cargo.toml depencies:
```
openssl = { version = "0.10", features = ["vendored"] }
```

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -349,6 +352,16 @@ for direct integration with IDEs and tools. 'http' starts an HTTP server for net
...
         port: u16,
     },
+
+    #[command(
+        about = "Print version information.",
+        short_flag = 'V',
+        long_flag = "version"
+    )]
+    Version {
+        #[arg(short = 'd', long = "debug", default_value_t = false, hide = true)]
```
weibenz1: `default_value_t` is it correct?

also, does `hide = true` mean it's callable but just not shown when user do `--help`?

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -1,18 +1,24 @@
...
 
 /// Simplified service definition with fields removed
 #[derive(Debug, Clone, Serialize, Deserialize)]
 struct SimplifiedServiceDefinition {
     #[serde(skip_serializing_if = "Option::is_none")]
     version: Option<String>,
     metadata: ServiceMetadata,
-    operations: HashMap<String, SimplifiedOperation>,
-    shapes: HashMap<String, SimplifiedShape>,
+    operations: BTreeMap<String, SimplifiedOperation>,
```
weibenz1, line 20: not strongly against it, but this change has performance complications - `get` in BTreeMap is LogN TC.

Considering only the version/hash computing logic need keys ordered and it's rarely used, duplicate/copy the data might be an better option here.

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -49,6 +55,72 @@ struct ShapeReference {
...
+    pub git_commit_hash: String,
+    /// the git tag of boto3/botocore, returned on calls to iam-policy-autopilot --version --debug
+    #[serde(rename = "gitTag")]
+    pub git_tag: Option<String>,
+    /// the sha hash of boto3/botocore simplified models, returned on calls to iam-policy-autopilot --version --debug
+    #[serde(rename = "dataHash")]
+    pub data_hash: String,
+}
+
+impl GitSubmoduleVersion {
```
weibenz1: lets add more comments/explanations for this struct

iam-policy-autopilot-cli/src/main.rs
hwei0: > also, does hide = true mean it's callable but just not shown when user do --help?

yes, think so, here's the output of --help:
```
./target/release/iam-policy-autopilot --version --help
Print version information.

Usage: iam-policy-autopilot {version|--version|-V}

Options:
  -h, --help  Print help
```

> default_value_t is it correct?

yes, think so, c.f. https://github.com/clap-rs/clap/discussions/3640

also here's the output if the `default_value_t` is set as `true`; the full boto3/botocore info is printed even if the --debug flag is omitted, which isn't desired:

```
./target/release/iam-policy-autopilot --version       
0.1.2
boto3 version: commit_id=b4d31fc1798a1cfa269433e8e8c9a7516f276896, commit_tag=1.42.13, data_hash=SHA256:8309474f524e2f3de2f90ada66f1c5c887358cc503846073663ba311783ff495
botocore version: commit_id=7c770fb210913ff9fa9b35470161c3278786085b, commit_tag=None, data_hash=SHA256:615e77e2c8c1dd6fd471416d712f4fe5a8e66b45150dc5d4bdc192736a8a42f1
```

iam-policy-autopilot-policy-generation/build.rs
hwei0: ok i'll change so that sorting happens only during version computation. 

But given that this logic is only within `build.rs` and not the core CLI logic, isn't the whole point of SimplifiedShape and SimplifiedServiceDefinition within `build.rs` just for a one-time construction + copying of data to the `build` dir? With BTree map we have `N` one-time additions, each of complexity `logN` (so total complexity `O(N log N)`). Without BTreeMap, we have `N` one-time O(1) additions, but a final `N log N` sorting step (so total complexity also `O(N log N)`). So isn't the complexity the same, either with or without BTreeMap?

iam-policy-autopilot-policy-generation/build.rs
hwei0: will do :)

iam-policy-autopilot-policy-generation/build.rs
weibenz1: I missed it's in build.rs. If it's only during build time, then BTree is fine. I mistakenly thought we are using BTree during policy generation.

hwei0: regardless i'll still rewrite using a `serialize_with` like this: https://stackoverflow.com/questions/42723065/how-to-sort-hashmap-keys-when-serializing-with-serde

```
use serde::{Deserialize, Serialize, Serializer}; // 1.0.106
use serde_json; // 1.0.52
use std::collections::{BTreeMap, HashMap};

#[derive(Serialize, Deserialize, Default)]
struct MyStruct {
    #[serde(serialize_with = "ordered_map")]
    map: HashMap<String, String>,
}

/// For use with serde's [serialize_with] attribute
fn ordered_map<S, K: Ord + Serialize, V: Serialize>(
    value: &HashMap<K, V>,
    serializer: S,
) -> Result<S::Ok, S::Error>
where
    S: Serializer,
{
    let ordered: BTreeMap<_, _> = value.iter().collect();
    ordered.serialize(serializer)
}

fn main() {
    let mut m = MyStruct::default();
    m.map.insert("gamma".into(), "3".into());
    m.map.insert("alpha".into(), "1".into());
    m.map.insert("beta".into(), "2".into());

    println!("{}", serde_json::to_string_pretty(&m).unwrap());
}

```

iam-policy-autopilot-cli/src/main.rs
weibenz1: yeah, I didn't mean to question the intention for setting the default value, just wondering if the syntax was correct.

iam-policy-autopilot-policy-generation/build.rs
weibenz1: why the rewrite if BTree already serves the purpose? Also, would we run into any scenarios that the ordering of values messes up our hash computation?

hwei0: ok fair, I reset to BTreeMap.

Yeah the ordering gets messed up a lot, without sorting the entries in those service definition structs I got different hashes on every re-run, when I was writing the code. I changed to BTree to fix that, since it makes each resulting service definition's serialized JSON get produced deterministically. I also made `sha2sum_recursive` compute the hash by sorting the paths in the directories it recurses over, since without that the sha2 computation was also non-deterministic.

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -49,6 +55,72 @@ struct ShapeReference {
...
+        }
+    }
+
+    fn sha2sum_recursive(cwd: &Path, root: &Path) -> Result<Digest, Box<dyn std::error::Error>> {
+        let mut hash_table: BTreeMap<RelativePathBuf, Digest> = BTreeMap::new();
+
+        let mut dir_entry_list = fs::read_dir(cwd)?
+            .map(|res| res.map(|e| e.path()))
+            .collect::<Result<Vec<_>, io::Error>>()?;
+        dir_entry_list.sort();
```
weibenz1, line 89: Q: why is this necessary?

weibenz1 (approved):

iam-policy-autopilot-policy-generation/build.rs
hwei0: The sha computation recurses the contents of the simplified boto3 and botocore subdirectories, and the ordering of this recursion affects the computed checksum. So the ordering of the recursion needs to be deterministic to make the hash deterministic.

Cargo.toml
mschlaipfer: By how much does that increase the binary size?

mschlaipfer: Not sure if there's a better solution, but if the size increase is significant, we should look into it more.

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -349,6 +352,16 @@ for direct integration with IDEs and tools. 'http' starts an HTTP server for net
...
+        debug: bool,
```
mschlaipfer: nit: `verbose` would be a more appropriate name?

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -349,6 +352,16 @@ for direct integration with IDEs and tools. 'http' starts an HTTP server for net
 Only used when --transport=http. The server will bind to 127.0.0.1 (localhost) on the specified port.")]
         port: u16,
     },
+
+    #[command(
+        about = "Print version information.",
+        short_flag = 'V',
+        long_flag = "version"
```
mschlaipfer, line 359: These `_flag` parameters make it so it's called as `--version` rather than `iam-policy-autopilot version`?

iam-policy-autopilot-policy-generation/resources/config/sdks/boto3
mschlaipfer, line 1: Does this update need to be part of this PR?

iam-policy-autopilot-policy-generation/Cargo.toml
```diff
@@ -27,6 +27,11 @@ serde_json.workspace = true
 tokio.workspace = true
 async-trait.workspace = true
 strsim.workspace = true
+aws-lc-rs.workspace = true
+git2.workspace = true
+relative-path.workspace = true
+openssl.workspace = true
```
mschlaipfer: Are these needed as non-build dependencies?

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -49,6 +55,72 @@ struct ShapeReference {
...
+
+        for entry_path in dir_entry_list {
+            let relt_path = entry_path.clone().relative_to(root)?;
+            if entry_path.is_dir() {
+                hash_table.insert(
+                    relt_path.clone(),
```
mschlaipfer: Is this `clone` needed (same question for `else`)?

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -49,6 +55,72 @@ struct ShapeReference {
...
+                );
+            } else {
+                let mut sha2_context = Context::new(&SHA256);
+                sha2_context.update(&fs::read(entry_path)?);
+                hash_table.insert(relt_path.clone(), sha2_context.finish());
+            }
+        }
+
+        let mut sha2_context = Context::new(&SHA256);
+        for entry in hash_table {
```
mschlaipfer:
```suggestion
        for (path, digest) in hash_table {
```

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -49,6 +55,72 @@ struct ShapeReference {
...
+                "{:?}",
+                Self::sha2sum_recursive(data_path, data_path).unwrap_or_else(|_| panic!(
+                    "Failed to compute checksum over data at path {:?}",
+                    data_path
+                ))
+            ),
+        }
+    }
+
+    fn sha2sum_recursive(cwd: &Path, root: &Path) -> Result<Digest, Box<dyn std::error::Error>> {
```
mschlaipfer, line 83: Could you add a comment how we do the hashing? Something like: "Each directory path is hashed, each file name and each file's content make up the hash sum."

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -49,6 +55,72 @@ struct ShapeReference {
     shape: String,
 }
 
+/// This must be an exact copy of GitSubmoduleMetadata in model.rs. This struct contains the boto3/botocore submodule version info; it then gets serialized into a JSON file, which is then referenced via the GitSubmoduleVersionInfoRaw RustEmbed struct in embedded_data.rs, so that this version info can get read when --version in the CLI is called.
```
mschlaipfer: How about defining this type in a separate file and using the `include!` macro in both places?

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -49,6 +55,72 @@ struct ShapeReference {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct GitSubmoduleVersion {
```
mschlaipfer: For consistency across the codebase, can we just use `#[serde(rename_all = "PascalCase")]`?

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -417,3 +522,29 @@ fn process_boto3_service_version(
 
     Ok(has_resources_file)
 }
+
+fn get_repository_tag(repo: &Repository) -> Result<Option<String>, Box<dyn std::error::Error>> {
+    // we want to do this: git describe --exact-match --tags
```
mschlaipfer: Are we doing it? :) Then I'd suggest making this a comment on the function like `// Performs git describe --exact-match --tags`

iam-policy-autopilot-cli/src/commands.rs
```diff
@@ -4,10 +4,10 @@
 
 use crate::{output, types::ExitCode};
 use iam_policy_autopilot_access_denied::{ApplyError, ApplyOptions, DenialType};
-
 fn is_tty() -> bool {
     atty::is(atty::Stream::Stdin) && atty::is(atty::Stream::Stderr)
 }
+use clap::crate_version;
```
mschlaipfer: Let's move this `use` to the top of the file.

mschlaipfer (changes requested):

iam-policy-autopilot-cli/src/main.rs
hwei0: I think `iam-policy-autopilot --version` was already originally the way to call the version command, and just using `iam-policy-autopilot version` also did not work before, if i'm not mistaken:
```
hbwei@6c7e67c57386 iam-policy-autopilot-awslabs % ./target/debug/iam-policy-autopilot --version        
iam-policy-autopilot 0.1.2

hbwei@6c7e67c57386 iam-policy-autopilot-awslabs % ./target/debug/iam-policy-autopilot version   
error: unrecognized subcommand 'version'

Usage: iam-policy-autopilot <COMMAND>

For more information, try '--help'.
```

iam-policy-autopilot-cli/src/main.rs
hwei0: fair enough, will change!

iam-policy-autopilot-policy-generation/resources/config/sdks/boto3
hwei0: no it doesnt, i'll restore the old one

iam-policy-autopilot-policy-generation/Cargo.toml
hwei0: you're right, they're not needed!

iam-policy-autopilot-policy-generation/build.rs
hwei0: not needed, will remove!

iam-policy-autopilot-policy-generation/build.rs
hwei0: will add!

iam-policy-autopilot-policy-generation/build.rs
hwei0: this is cool, didn't know about this. will do that, thanks!

iam-policy-autopilot-policy-generation/build.rs
hwei0: yeah what's being performed there, is intended to do the equivalent of `git describe --exact-match --tags`. Will update the comment!

iam-policy-autopilot-cli/src/commands.rs
weibenz1: im surprised that cargo fmt and clippy didn't catch it

iam-policy-autopilot-cli/src/main.rs
mschlaipfer: Yes, I just want to make sure it's `iam-policy-autopilot --version`, which you confirmed

iam-policy-autopilot-policy-generation/src/shared_submodule_model.rs
```diff
@@ -0,0 +1,12 @@
+
+/// Exposes git version and commit hash for boto3 and botocore
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct GitSubmoduleMetadata {
+    /// the commit of boto3/botocore, returned on calls to iam-policy-autopilot --version --debug
```
mschlaipfer: NP: outdated comment `--debug` (also below)

iam-policy-autopilot-policy-generation/src/shared_submodule_model.rs
```diff
@@ -0,0 +1,12 @@
+
```
mschlaipfer: not sure why there's an empty line, could you also add a comment at the top of this file that explains why we define this separately and `include!` it?

Cargo.toml
hwei0: oh lol the binary just stays the same size: 
```
[current release]
hbwei@6c7e67c57386 iam-policy-autopilot-awslabs % du -sh ./target/release/iam-policy-autopilot
 53M    ./target/release/iam-policy-autopilot

[this PR changes]
hbwei@6c7e67c57386 iam-policy-autopilot % du -sh target/release/iam-policy-autopilot
 52M    target/release/iam-policy-autopilot
```

maybe it's because I just added `openssl` to the workspace dependencies, which don't get packaged into the binary

weibenz1 (approved):

Issue #92, opened 2025-12-19
Introduce a disable interaction flag to disable any put policy operations as a security guardrail


karanjitsingh:
Introduce a --read-only flag to the iam-policy-autopilot MCP server. While system prompts can discourage policy modifications, they are non-deterministic and can be bypassed. This flag will provide a deterministic security guardrail by explicitly disabling all Put and write-related operations to AWS IAM policies regardless of the model's intent.

jmolz:
Opened PR #227 to implement this: https://github.com/awslabs/iam-policy-autopilot/pull/227

The PR adds an MCP server --read-only flag, keeps policy generation tools available, and rejects the mutating fix_access_denied MCP tool before policy planning/apply when read-only mode is enabled.

PR #93, opened 2025-12-19
feat: read only mode as a security guardrail (#92)

karanjitsingh:
*Issue #, if available:* #92

*Description of changes:*
Introduces `--read-only` flag as a security guardrail. currently `fix_access_denied` is the only tool that mutates policies in customer's account, we should simply disable this tool if read-only flag is set, since `generate_policy_for_access_denied` already exists, no need for custom logic within `fix_access_denied`


iam-policy-autopilot-mcp-server/src/mcp.rs
```diff
@@ -166,14 +171,52 @@ impl ServerHandler for IamAutoPilotMcpServer {
...
+        _: Option<PaginatedRequestParam>,
+        _: RequestContext<RoleServer>,
+    ) -> Result<ListToolsResult, McpError> {
+        let all_tools = self.tool_router.list_all();
+
+        // Filter out fix_access_denied tool when in read-only mode
+        let tools: Vec<Tool> = if self.read_only {
+            all_tools
+                .into_iter()
+                .filter(|tool| tool.name.as_ref() != "fix_access_denied")
```
mschlaipfer: Instead of filtering by name, would it make sense for the tools to have a property against we filter?

iam-policy-autopilot-mcp-server/src/lib.rs
```diff
@@ -24,7 +24,7 @@ impl Display for McpTransport {
     }
 }
 
-pub async fn start_mcp_server(transport: McpTransport, port: u16) -> Result<()> {
+pub async fn start_mcp_server(transport: McpTransport, port: u16, read_only: bool) -> Result<()> {
```
mschlaipfer: Instead of passing individual booleans, let's introduce a configuration object.

PR #97, opened 2026-01-12
refactor: consolidate waiter and paginator structs

adpaco-aws:
*Issue #, if available:* N/A

*Description of changes:* Consolidated waiter and paginator extraction structs from Python and Go extractors into a unified shared module. This eliminates code duplication and creates a consistent API across languages.

As part of this refactoring, we have replaced line-related fields with `start_position` and `end_position` tuple fields for better precision. I noticed that we could go even further and have `WaiterCreationInfo` and `PaginatorCreationInfo` become a single struct `UtilityCreationInfo` but I decided against it to have the code be clearer about what it's doing.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/extraction/shared/mod.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/shared/mod.rs
@@ -1,3 +1,3 @@
 pub mod extraction_utils;
 
-pub use extraction_utils::*;
+pub(crate) use extraction_utils::*;


iam-policy-autopilot-policy-generation/src/extraction/go/paginator_extractor.rs
```diff
@@ -170,13 +140,17 @@ impl<'a> GoPaginatorExtractor<'a> {
             .and_then(|s| s.strip_suffix("Paginator"));
 
         if let Some(operation_name) = operation_name {
-            return Some(PaginatorInfo {
+            let node = node_match.get_node();
+            let location = Location::from_node(file_path.to_path_buf(), &node);
+            let expr = node_match.text().to_string();
```
adpaco: These assignments are done outside the constructor because the previous positions (i.e., `start_position` and similar fields) were assigned like this. Happy to do this into the construction below as written in #84 .

iam-policy-autopilot-policy-generation/src/extraction/shared/extraction_utils.rs
```diff
@@ -0,0 +1,91 @@
+use crate::extraction::core::Parameter;
+use crate::Location;
+
+/// Information about a discovered waiter creation call (get_waiter in Python, NewXxxWaiter in Go)
+#[derive(Debug, Clone)]
+pub struct WaiterCreationInfo {
```
mschlaipfer: I think these should all be internal `pub(crate)`?

iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
```diff
@@ -7,38 +7,18 @@
...
-
-impl WaiterInfo {
-    pub(crate) fn start_line(&self) -> usize {
-        self.location.start_line()
-    }
-}
-
 // TODO: This should be refactored at a higher level, so this type can be removed.
 // See https://github.com/awslabs/iam-policy-autopilot/issues/88.
 enum CallInfo<'a> {
```
mschlaipfer, line 19: Would it make sense to remove/move this type now that we are consolidating this? Could be done in a separate PR.

iam-policy-autopilot-policy-generation/src/extraction/shared/extraction_utils.rs
adpaco: Yes, good catch.

iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
adpaco: I'd rather do it in a separate PR.

PR #98, opened 2026-01-12
ci: remove commit message validation and improve triggering rules

weibenz1:
*Issue #, if available:* 

N/A

*Description of changes:*
This PR introduces two changes to our PR checks:
1. removed commit message validation given we rely on rulesets now which has better enforcement
2. Improved triggering rules for our pr-checks.yml so we only build & compile & test if there's rust related changes (e.g. readme changes will not consume our github action usage)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,6 +58,9 @@ We follow [Conventional Commits](https://www.conventionalcommits.org/) for commi
 - `refactor:` - Code refactoring without changing functionality
 - `perf:` - Performance improvements
 - `ci:` - CI/CD configuration changes
+- `build:` - Compilation/build related changes
+- `style:` - Formatting changes
+- `revert:` - Reverted commits
 
 **Breaking Changes:**
 
@@ -65,8 +68,6 @@ For breaking changes, append `!` after the type or add `BREAKING CHANGE:` in the
 
 ```bash
 feat!: change CLI argument format for generate-policies command
-
-BREAKING CHANGE: --region flag is now required instead of optional
 ```
 
 **Examples:**


.github/workflows/pr-checks.yml
```diff
@@ -8,51 +8,28 @@ permissions:
...
-            echo "Valid types: feat, fix, docs, chore, test, refactor, perf, ci, build, style, revert"
-            echo "For breaking changes, add '!' after type: feat!: breaking change"
-            echo ""
-            echo "See CONTRIBUTING.md for more details."
-            exit 1
-          fi
+          filters: |
+            rust:
+              - '**/*.rs'
+              - '**/*.toml'
```
mschlaipfer, line 24: Are there relevant `.toml` files other than `Cargo.toml`?

.github/workflows/pr-checks.yml
```diff
@@ -8,51 +8,28 @@ permissions:
...
+              - '**/Cargo.lock'
+              - '**/rust-toolchain'
+              - '**/.cargo/**'
```
mschlaipfer, line 27: What about changes to `.yml` files like this very workflow?

.github/workflows/pr-checks.yml
```diff
@@ -8,51 +8,28 @@ permissions:
...
-            echo "Commit messages must follow Conventional Commits format:"
-            echo "  <type>: <description>"
-            echo ""
-            echo "Valid types: feat, fix, docs, chore, test, refactor, perf, ci, build, style, revert"
-            echo "For breaking changes, add '!' after type: feat!: breaking change"
-            echo ""
-            echo "See CONTRIBUTING.md for more details."
-            exit 1
-          fi
+          filters: |
```
mschlaipfer, line 21: Rather than allow-listing what triggers the checks, could we deny-list which files don't need to trigger them (such as `.md`)?

.github/workflows/pr-checks.yml
weibenz1: yes, for example, https://github.com/awslabs/iam-policy-autopilot/blob/main/dylint.toml - custom linting implemented by @adpaco-aws, and https://github.com/awslabs/iam-policy-autopilot/blob/main/pyproject.toml for packaging the pypi binaries

.github/workflows/pr-checks.yml
weibenz1: changes to yml files will always be picked by github to run validation

mschlaipfer (approved):

.github/workflows/pr-checks.yml
weibenz1: I've thought about it, but it needs more hardcoding, like LICENSE files, power files etc, and any new things.
Let's keep it and revise if needed in the future?

mschlaipfer: Right, but it fails in the safe way, we'd run checks more often than needed. Currently, we might miss running the checks.

PR #103, opened 2026-01-14
fix: support eu sovereign cloud

C85297:
Currently, `AwsContext` uses some heuristics to guess the corresponding partition for a given region. This is not ideal as it requires manual fixes when new partitions are created by AWS. The new EU Sovereign Cloud partition is not recognised by the heuristics.

I updated `AwsContext` to pull in botocore's `partitions.json` file. This file provides a region regex for each partition, which we can use to check if a region is in a given partition. This ensures we can automatically support new partitions without any work as soon as they are added to botocore.

Note: the existing code would assume unknown regions are in the `aws` partition. I would suggest this is not sensible behaviour - if we have the `aws` partition's region regex and know it doesn't match, then it is more likely that the region is either invalid or in a partition we are not aware of yet. In both those cases, I would say it would be a better idea to exit with an error, which I've implemented. 

The code does not cache the parsed partitions data, so multiple invocations of `AwsContext::new` will be unnecessarily expensive. Since `AwsContext::new` is only invoked once during the CLI's execution, and no more than once during an MCP call, I decided to keep the code simple. I can change this if you'd prefer.


iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -109,34 +111,42 @@ mod tests {
...
 
-        let ctx = AwsContext::new("ap-southeast-1".to_string(), "123456789012".to_string());
+        let ctx =
+            AwsContext::new("ap-southeast-1".to_string(), "123456789012".to_string()).unwrap();
         assert_eq!(ctx.partition, "aws");
+    }
 
-        // Test wildcard
-        let ctx = AwsContext::new("*".to_string(), "*".to_string());
-        assert_eq!(ctx.partition, "aws");
```
adpaco, line 143: Why are we removing this test?

iam-policy-autopilot-policy-generation/build.rs
```diff
@@ -55,6 +55,19 @@ struct ShapeReference {
     shape: String,
 }
 
+/// Simplified partitions definition with fields removed
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct SimplifiedPartitionsDefinition {
+    partitions: BTreeMap<String, SimplifiedPartition>,
+}
```
adpaco: There is no need for `SimplifiedPartition`, we could directly use `String`. Are you doing this to achieve a specific serialization format?

iam-policy-autopilot-policy-generation/src/api/model.rs
C85297: Ah, sorry, meant to mention this in the pull request but I forgot. I couldn't work out the intended use case this test case was supporting. It seemed that a user could provide the option `--region *` and autopilot would assume `aws` partition. This isn't actually documented anywhere as far as I can tell, and doesn't make much sense to me.

If this is intended I could try and restore this behaviour, but it might make more sense to instead set the partition to `*` so the outputted policies are generic over any partition. However, this approach would get complicated or impossible when applied to service principals.

Would appreciate further guidance on what use case this is intended for, as then I could also update the docs to highlight that!

iam-policy-autopilot-policy-generation/build.rs
C85297: @adpaco-aws I did this to allow easily consuming more partition fields other than just `regionRegex` if needed in the future, however I think you are right, so I've made it just a direct Regex for simplicity. Shouldn't be too hard to change it back to a struct if needed in the future.

C85297:
> Thanks! In my opinion, it would be nice if we cached the partitions data. I don't think it'll make the code much more complicated and it'd be ready to call more than once if we need to do that in the future. At the very least, I'd open an issue and add a comment to note we're missing caching.

@adpaco-aws Updated with caching implemented within `get_partition` using a `LazyLock`. Its a bit complicated due to `ExtractorError` not being able to be cloneable, but works well.

iam-policy-autopilot-policy-generation/src/embedded_data.rs
```diff
@@ -241,6 +255,45 @@ impl Boto3Data {
 pub(crate) struct BotocoreData;
 
 impl BotocoreData {
+    /// Get a parsed partitions definition file
+    ///
+    /// # Returns
+    /// Parsed partitions definition or error if not found or parsing fails
+    pub(crate) fn get_partitions(
+    ) -> std::result::Result<&'static PartitionsDefinition, &'static ExtractorError> {
+        static PARTITIONS: LazyLock<std::result::Result<PartitionsDefinition, ExtractorError>> =
```
adpaco, line 264: Do you mind moving the static declaration to the module level and using `OnceLock` instead for consistency with the rest of the code base?

iam-policy-autopilot-policy-generation/src/api/model.rs
mluttrell: I think you're spot on for what this use case was designed for. I agree that we should set the partition value to `*` if the user has essentially said that they do not need a region nor an account value specified.  The value of providing a partition value is fairly small, regardless. 

I think that a wildcarded partition value should work in every case I can think of.

adpaco: Sorry for the delay in getting back to you @C85297 - I wanted to get feedback from other people on this use case. Please follow @mluttrell 's recommendation for resolution.

mschlaipfer:
No worries, just wanted to check in. Thanks for getting back so quickly!

iam-policy-autopilot-policy-generation/src/embedded_data.rs
C85297: @adpaco-aws agree with generally looking to keep consistency, but in this case OnceLock and LazyLock have different behaviours and functionality, and in this case using OnceLock would increase verbosity and complexity, and could lead to future contributors making mistakes.

OnceLock requires the initialisation code to be provided everywhere you get the value. For a variable for which there is only one way value would be initialised, this means that whenever a new retrieval of the value is added, all the code to initialise the value must be duplicated. This introduces unnecessary verbosity and complexity and could lead to future errors, if someone accidentally modifies only one of the initialisers, not realising that there are others.

LazyLock is better suited for a situation like this, where there is only one correct way to initialise the value, and it also provides better semantics, as no ‘get’ call is required - the variable can be used as if it is the type of the initialised value.

The OnceLock docs also recommend using a LazyLock in this type of scenario: https://doc.rust-lang.org/std/sync/struct.OnceLock.html#:~:text=In%20many%20simple,call%20LazyLock::new(%7C%7C%20...).

iam-policy-autopilot-policy-generation/src/api/model.rs
C85297: @mluttrell Happy to implement this behaviour, however just to make you aware I believe we will have to change how FAS maps work in this project. Currently the service principal names are hardcoded for the AWS partition. It should be easy enough to change this, and make dynamically fetch the correct ones based on the partition and region. But this is one thing that cannot be wild carded - so if someone passes in `—region *` the outputted policies would only work for regions in the AWS partition, which may lead to significant confusion as public use of the EU Sovereign Cloud partition increases.

mluttrell: We probably will need a partitional FAS mapping at some point in the future, but generally speaking, service principal names are the same across partitions.  Happy to be told you've found some exceptions to that, though!

iam-policy-autopilot-policy-generation/src/embedded_data.rs
adpaco: Thanks for the detailed explanation! I wasn't aware of this subtlety between `OnceLock` and `LazyLock`. You're right that `LazyLock` is better suited here since there's only one correct initialization path. Fine to keep it as is.

iam-policy-autopilot-policy-generation/src/api/model.rs
C85297: @mluttrell apologies, I've done a bit more digging and you're correct, looks like all partitions now use `amazonaws.com`: https://github.com/aws/aws-cdk/blob/f8e38638bb5be02bf006888c5d815570c4933c95/packages/aws-cdk-lib/region-info/lib/default.ts#L27

C85297: @mluttrell @adpaco-aws I've added back in this behaviour, and I've also updated the unit tests for policy-generation and the CLI integration tests not only to test wildcard region behaviour, but also wildcard account ID behaviour.

adpaco-aws (approved):

PR #104, opened 2026-01-14
fix: non-commercial partition and region support in access-denied

C85297:
Autopilot fix-access-denied currently does not support non-commercial AWS partitions and regions, such as US Gov Cloud and the EU Sovereign Cloud. 

I've updated the code to assume a partition beginning with `aws-` is a valid non-commercial AWS partition, and to handle these as far as possible. In cases where we are unable to detect what the relevant partition is, I've updated the code to output resources with a wildcard `*` in the partition part of the ARN. This will ensure the policy changes work across any partition. I've also updated and expanded the test coverage in this area.

Note: I updated the use of `OnceLock` to a `LazyLock`. This reduces code duplication as the initialisation code only needs to be specified once.

### DIFF
--- a/iam-policy-autopilot-access-denied/src/parsing/utils.rs
+++ b/iam-policy-autopilot-access-denied/src/parsing/utils.rs
@@ -1,10 +1,16 @@
 //! Utility functions for string-based AccessDenied message parsing.
 
 use regex::Regex;
-use std::sync::OnceLock;
+use std::sync::LazyLock;
 
 /// Compiled regex for ARN validation - only place we use regex in parsing
-static ARN_PATTERN: OnceLock<Regex> = OnceLock::new();
+static ARN_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"arn:aws(?:-[a-z]+)*:[a-zA-Z0-9\-]+:[a-zA-Z0-9\-]*:\d{12}:[^\s\x22]+").unwrap()
+});
+
+pub fn is_account_id(value: &str) -> bool {
+    value.len() == 12 && value.chars().all(|c| c.is_ascii_digit())
+}
 
 /// Validates AWS ARN format using basic pattern matching.
 pub fn is_arn(value: &str) -> bool {
@@ -12,7 +18,7 @@ pub fn is_arn(value: &str) -> bool {
         return false;
     }
     let value = value.trim();
-    if !value.starts_with("arn:aws:") {
+    if !value.starts_with("arn:aws:") && !value.starts_with("arn:aws-") {
         return false;
     }
 
@@ -21,11 +27,15 @@ pub fn is_arn(value: &str) -> bool {
         return false;
     }
 
+    let partition = parts[1];
     let service = parts[2];
     let region = parts[3];
     let account = parts[4];
     let resource = parts[5];
 
+    if partition != "aws" && !partition.chars().all(|c| c.is_alphanumeric() || c == '-') {
+        return false;
+    }
     if service.is_empty() || !service.chars().all(|c| c.is_alphanumeric() || c == '-') {
         return false;
     }
@@ -34,16 +44,13 @@ pub fn is_arn(value: &str) -> bool {
     }
 
     match service {
-        "s3" => region.is_empty() && account.is_empty(),
-        "iam" => {
-            region.is_empty()
-                && (account.is_empty()
-                    || (account.len() == 12 && account.chars().all(|c| c.is_ascii_digit())))
+        "s3" => {
+            (region.is_empty() && account.is_empty())
+                || (region.is_empty() && !account.is_empty() && is_account_id(account))
+                || (!region.is_empty() && !account.is_empty() && is_account_id(account))
         }
-        _ => {
-            account.is_empty()
-                || (account.len() == 12 && account.chars().all(|c| c.is_ascii_digit()))
-        }
+        "iam" => region.is_empty() && (account == "aws" || is_account_id(account)),
+        _ => account.is_empty() || is_account_id(account),
     }
 }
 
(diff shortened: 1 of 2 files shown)


iam-policy-autopilot-access-denied/src/parsing/utils.rs
```diff
@@ -21,11 +27,15 @@ pub fn is_arn(value: &str) -> bool {
         return false;
     }
 
+    let partition = parts[1];
     let service = parts[2];
     let region = parts[3];
     let account = parts[4];
     let resource = parts[5];
 
+    if partition != "aws" && !partition.chars().all(|c| c.is_alphanumeric() || c == '-') {
```
weibenz1, line 36: it will accept invalid partitions which makes the access denial fix never work.

Can we define a list of supported partitions and validate against that?

iam-policy-autopilot-access-denied/src/parsing/utils.rs
```diff
@@ -284,19 +285,30 @@ pub fn normalize_s3_resource(action: &str, resource: &str) -> String {
...
-    let after_prefix = &resource[prefix.len()..];
-
     // Find the first '/' which indicates this is an object ARN
-    if let Some(slash_pos) = after_prefix.find('/') {
-        let bucket_name = &after_prefix[..slash_pos];
+    if let Some(slash_pos) = trimmed.find('/') {
+        let bucket_name = &trimmed[..slash_pos];
         // Return bucket wildcard pattern
-        format!("arn:aws:s3:::{}/*", bucket_name)
+        format!("arn:{partition}:s3:::{bucket_name}/*")
```
weibenz1, line 311: Am I right to think that we might convert `bucketName/prefix/*` to `bucketName/*`? which will open up the access - should we capture the entire section surrounded by `/`?

weibenz1 (commented):
Thanks for the contribution!

iam-policy-autopilot-access-denied/src/parsing/utils.rs
C85297: @weibenz1 could you clarify what you mean by "the access denial fix never work"?

With this code, partitions which are unknown will be presumed to be valid and the code will correctly operate on this assumption. For example, the error:
```
User: arn:aws-not-a-partition:iam::123456789012:user/testuser is not authorized to perform: s3:GetObject on resource: arn:aws-not-a-partition:s3:::my-bucket/my-key
```
Will successfully output a plan to correctly add permissions scoped to that unknown partition:
```
PlanResult { diagnosis: ParsedDenial { principal_arn: "arn:aws-not-a-partition:iam::123456789012:user/testuser", action: "s3:GetObject", resource: "arn:aws-not-a-partition:s3:::my-bucket/*", denial_type: Other }, actions: ["s3:GetObject"], policy: PolicyDocument { id: Some("IamPolicyAutopilot"), version: "2012-10-17", statement: [Statement { sid: "IamPolicyAutopilotAllow", effect: "Allow", action: Single("s3:GetObject"), resource: "arn:aws-not-a-partition:s3:::my-bucket/*" }] } }
```

The problem here is that, with the data available to the `access-denied` crate, it is impossible to distinguish between an unknown and an actually invalid region (other than hardcoding the current partitions, which would not be ideal). If the user has an error, with an ARN containing a partition that appears to be in the right format, I thought it reasonable to proceed on the basis that this is a valid partition, and that the output plan should be scoped to that partition.

An alternative path I could take would be to use the data in botocore's partitions.json, however that would require wider refactoring of the codebase as currently that data is only available in the policy-generation crate, and not the access-denied crate.

iam-policy-autopilot-access-denied/src/parsing/utils.rs
C85297: @weibenz1 Yep, thats the original behaviour which I maintained: https://github.com/C85297/iam-policy-autopilot/blob/40309f4e259e9430d4577382f02b2d6a28ae4e36/iam-policy-autopilot-access-denied/src/parsing/utils.rs#L374

Sounds like a good area to improve, though I would suggest we could look at that in a further PR?

weibenz1: Yes, that can happen in a separate PR.

Created #107 for this

iam-policy-autopilot-access-denied/src/parsing/utils.rs
weibenz1: When a partition like `aws-not-a-partition` is provided, the proposed policies will never help with resolve the access denied error.

I think the bigger question is how to make IPA smarter in terms distinguish valid vs invalid partitions/regions etc - we can wait and see if users run into this often enough to deserve a tracing issue.

For now, I think what you propose here is fine - we assume user copies the exact error message - if they copied invalid/wrong error message, IPA returning invalid policy should be okay.

iam-policy-autopilot-access-denied/src/parsing/utils.rs
```diff
@@ -34,16 +44,13 @@ pub fn is_arn(value: &str) -> bool {
...
-        }
-        _ => {
-            account.is_empty()
-                || (account.len() == 12 && account.chars().all(|c| c.is_ascii_digit()))
+        "s3" => {
+            (region.is_empty() && account.is_empty())
+                || (region.is_empty() && !account.is_empty() && is_account_id(account))
+                || (!region.is_empty() && !account.is_empty() && is_account_id(account))
         }
+        "iam" => region.is_empty() && (account == "aws" || is_account_id(account)),
```
mschlaipfer, line 52: Looks like `cloudfront` is also possible in the account part. https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html#migrate-from-oai-to-oac.

I'm wondering if we need this precision in ARN validation here or would checking for non-emptiness be enough?

C85297: 🤔 I would have thought non-emptiness would be enough...

mschlaipfer: Opened https://github.com/awslabs/iam-policy-autopilot/issues/118

Issue #101, opened 2026-01-13
`test_get_botocore_version_info_happy_path` needs `#[test]` annotation


mschlaipfer:
https://github.com/awslabs/iam-policy-autopilot/blob/40309f4e259e9430d4577382f02b2d6a28ae4e36/iam-policy-autopilot-policy-generation/src/embedded_data.rs#L560

adpaco-aws:
This was resolved in #114, feel free to re-open if that's not the case.

PR #114, opened 2026-01-22
chore: infer lints from workspace and fix errors

mschlaipfer:
*Issue #, if available:* #101

*Description of changes:*

The dead code issue in #101 exhibited a deeper issue: we did not infer the workspace lint configuration in the crates. Fixing that (and adding the `dead_code = "deny"` setting) exposed the use of `unsafe_code` in the MCP server crate used for mocking. I refactored the code to get rid of `unsafe` and keep the `unsafe_code = "forbid"` rather than using `unsafe_code = "deny"` and introducing an `allow(unsafe_code)` override (`forbid` cannot be overridden).

I'd argue that `unsafe` code within `cfg(test)` is okay, but it seems like cargo does not allow to set `deny` only for `cfg(test)` modules.

Inferring linter settings in the crates exposed more issues. I have added explicit allow overrides for all failures and fixed the `unwrap_used` ones, sometimes just via changing the `unwrap` to an `expect`. We should step by step fix these `clippy` warnings.

### DIFF
--- a/iam-policy-autopilot-access-denied/src/aws/policy_naming.rs
+++ b/iam-policy-autopilot-access-denied/src/aws/policy_naming.rs
@@ -48,6 +48,7 @@ pub fn build_canonical_policy_name(_kind: &PrincipalKind, name: &str) -> String 
 
 /// Generate unique Sid with format IamPolicyAutopilot{Service}{Action}{YYYYMMDD}
 /// Handles collision detection by appending counter (2, 3, etc.)
+#[allow(unknown_lints, convert_case_pascal)]
 pub fn build_statement_sid(action: &str, date: &str, existing_sids: &[String]) -> String {
     // Falls back to "Unknown" service if format is invalid
     let parts: Vec<&str> = action.split(':').collect();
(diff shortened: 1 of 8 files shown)


mschlaipfer:
If reviewers agree with adding `"allow"` overrides for clippy findings in this PR, I'll open a separate issue to fix the clippy findings.

iam-policy-autopilot-policy-generation/src/extraction/javascript/scanner.rs
```diff
@@ -539,10 +536,9 @@ where
                 let method_name = method_node.text().to_string();
 
                 // Check if it's a known client variable
-                if client_variables.contains(&variable_name) {
-                    let (client_type, original_client_type, client_sublibrary) =
```
adpaco, line 543: The condition has been updated, how confident are we about this?

mschlaipfer: This essentially just checked for containment of `variable_name` in `keys()`: line 598 just below this in the diff. That's why the `get` with `unwrap` was okay.

Issue #99, opened 2026-01-13
Remove `name` field in `Resource`


mschlaipfer:
Called out in https://github.com/awslabs/iam-policy-autopilot/pull/84. The `name` field is not used so we should remove it.

mschlaipfer:
Decided to keep it, as the resource name/type might become useful in the future.

PR #115, opened 2026-01-22
refactor: remove name from Resource

mschlaipfer:
*Issue #, if available:* #99

*Description of changes:*
`name` in `Resource` is unused, this CR removes it.


iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -347,17 +347,15 @@ impl Action {
 /// metadata to provide complete resource information for IAM policies.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub(crate) struct Resource {
-    /// The resource type name (e.g., "bucket", "object", "*")
-    pub(crate) name: String,
```
adpaco, line 351: Do we even need a `struct` anymore? Could be just a `type`...

PR #122, opened 2026-01-26
feat: action filtering for `--explain`

mschlaipfer:
*Issue #, if available:*

*Description of changes:*

Make `--explain` take a list of actions, with wildcard support. Add a sentence about `--explain` to the `README`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Added
 
-- Add `--explain` feature to output the reasons for why an action has been added to the policy. The explanations allow to review the operations which static analysis extracted from source code, and to correct them using the `--service-hints` flag, if necessary.
+- Add `--explain` feature with action pattern filtering to output the reasons for why a actions were added to the policy. Supports wildcards (e.g., `--explain '*'` for all, `--explain 's3:*'` for S3 actions). The explanations allow to review the operations which static analysis extracted from source code, and to correct them using the `--service-hints` flag, if necessary.
 
 ## [0.1.2] - 2025-12-15
 


iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
```diff
@@ -8,11 +9,112 @@ use crate::{
...
+    let action_lower = action.to_lowercase();
+
+    // Split pattern by '*' to get literal parts that must appear in order
+    // e.g., "ec2:Describe*" → ["ec2:describe", ""]
+    // e.g., "*:Get*" → ["", ":get", ""]
+    let parts: Vec<&str> = pattern_lower.split('*').collect();
+
+    if parts.len() == 1 {
+        // No wildcard, exact match required
+        return action_lower == pattern_lower;
```
weibenz1, line 55: nit: shall we ignore spaces?

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
```diff
@@ -8,11 +9,112 @@ use crate::{
...
+    // After finding each part, we advance past it to enforce order
+    let mut remaining = action_lower.as_str();
+
+    for (i, part) in parts.iter().enumerate() {
+        // Skip empty parts (from consecutive wildcards or leading/trailing wildcards)
+        if part.is_empty() {
+            continue;
+        }
+
+        if i == 0 {
```
weibenz1, line 68: can we not convert * to a regex pattern, and do regex matching?

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -321,14 +321,22 @@ Use this flag to force fresh data retrieval on every run."
...
-        /// Generate explanations for why actions were added
+        /// Generate explanations for why actions were added, filtered to specific action patterns
         #[arg(
             long = "explain",
-            long_help = "When enabled, generates detailed explanations for why each IAM action \
-was added to the policy. Explanations include the initial operation with location information, FAS (https://docs.aws.amazon.com/IAM/latest/UserGuide/access_forward_access_sessions.html) expansion chains. The output format \
-may change in future versions."
+            num_args = 1..,
+            value_name = "ACTION_PATTERNS",
+            long_help = "Generates detailed explanations for why IAM actions were added to the policy. \
```
weibenz1, line 329: what happens if input is `--explain '   '`? (empty string)

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
mschlaipfer: That was my initial thought as well. Claude came up with this and convinced me that this works.

Regex would be more complicated because we'd need to ensure we escape correctly, but I could look into it if you have concerns about correctness of this approach.

iam-policy-autopilot-cli/src/main.rs
mschlaipfer: No action matches the empty string, so we'd filter all explanations and we wouldn't output an `"Explanations"` object.

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
mschlaipfer: I don't think so, also: spaces are currently used as separators between actions when multiple actions are passed to `--explain`.

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
mschlaipfer: I want to be clear that I don't want to support full regex, but just `*`

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -321,14 +321,22 @@ Use this flag to force fresh data retrieval on every run."
...
 
-        /// Generate explanations for why actions were added
+        /// Generate explanations for why actions were added, filtered to specific action patterns
         #[arg(
             long = "explain",
-            long_help = "When enabled, generates detailed explanations for why each IAM action \
-was added to the policy. Explanations include the initial operation with location information, FAS (https://docs.aws.amazon.com/IAM/latest/UserGuide/access_forward_access_sessions.html) expansion chains. The output format \
-may change in future versions."
+            num_args = 1..,
+            value_name = "ACTION_PATTERNS",
```
adpaco, line 328: This is technically a breaking change, using `--explain` without arguments used to work. We either fix it or call it out.

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
weibenz1: I'd still imagine we want to trim spaces after removing all the separators? But okay with this for now and see if customers want something else.

README.md
```diff
@@ -42,7 +42,7 @@ IAM Policy Autopilot stays up to date with the latest AWS services and features
 
 ### Review and refine policies generated by IAM Policy Autopilot
 
-IAM Policy Autopilot generates baseline policies to provide a starting point that you can refine as your application matures. Review the generated policies to ensure they align with your security requirements before deploying them.
+IAM Policy Autopilot generates baseline policies to provide a starting point that you can refine as your application matures. Review the generated policies to ensure they align with your security requirements before deploying them. Use the `--explain <action name>` feature to understand which operations led to an action being included in the generated policies.
```
adpaco: It'd be clearer if we explicitly mention action patterns and show an example. What about this?
```suggestion
IAM Policy Autopilot generates baseline policies to provide a starting point that you can refine as your application matures. Review the generated policies to ensure they align with your security requirements before deploying them. Use the `--explain` feature with action patterns (e.g., `--explain 's3:*'`) to understand which operations led to an action being included in the generated policies.
```

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
mschlaipfer: I should check how `clap` parses the arguments. Maybe I'm not following which spaces we'd trim.

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
```diff
@@ -8,11 +9,112 @@ use crate::{
...
+/// subsequent parts can only be found in what remains, enforcing left-to-right order.
+///
+/// Example: Pattern `"*item*get*"` vs Action `"s3:getitemversion"`
+/// - parts = `["", "item", "get", ""]`
+/// - Find `"item"` at position 6, remaining becomes `"version"`
+/// - Find `"get"` in `"version"` → NOT FOUND (even though "get" exists before "item")
+/// - Result: NO MATCH ✓
+fn action_matches_pattern(action: &str, pattern: &str) -> bool {
+    let pattern_lower = pattern.to_lowercase();
+    let action_lower = action.to_lowercase();
```
adpaco, line 46: What's the motivation for case insensitive?

iam-policy-autopilot-cli/src/main.rs
mschlaipfer: It's not been released. I'll update the changelog entry

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
mschlaipfer: Action matching in IAM policies is case-insensitive, so I thought it's natural to allow case-insensitive comparison here as well.

iam-policy-autopilot-cli/src/main.rs
adpaco: True, I thought it had been released for some reason.

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
adpaco: Makes sense, thanks.

PR #128, opened 2026-01-29
feat: variable type tracking for boto3 clients and resources

adpaco:
*Description of changes:*

## Summary

Adds variable type tracking to improve SDK method call extraction precision when boto3 clients and resources are passed across function boundaries. Resolves all review feedback from the initial submission.

## What's Tracked

- **Direct assignments**: `s3_client = boto3.client('s3')`, `dynamodb = boto3.resource('dynamodb')`
- **Multi-argument calls**: `boto3.client('s3', region_name='us-east-1')` (extra kwargs ignored)
- **Variable aliases**: `my_client = s3_client`
- **Function parameters**: Type inference from call sites (supports multiple types per parameter)
- **Keyword arguments at call sites**: `helper(client=s3)` matched by name, not position
- **Resource-derived variables**: `table = dynamodb.Table('users')`, `bucket = s3.Bucket('name')`
- **Python LEGB scoping**: Function-local variables and parameters correctly shadow module-level

### Direct Client Assignment

```python
import boto3
acm_client = boto3.client('acm')
# ListCertificates exists in acm, iot, AND transfer
response = acm_client.list_certificates()
```

**Before:** `PossibleServices: ["acm", "iot", "transfer"]` -> 3 policy statements
**After:** `PossibleServices: ["acm"]` -> 1 policy statement

### Function Parameter Tracking

```python
import boto3

def list_all_certificates(client):
    response = client.list_certificates()
    return response.get('CertificateSummaryList', [])

acm_client = boto3.client('acm')
certificates = list_all_certificates(acm_client)
```

**Before:** `["acm", "iot", "transfer"]`
**After:** `["acm"]`

## Testing

- **50 unit tests** covering all tracking patterns
- Tests use `rstest` for parameterized cases where appropriate
- Coverage includes: keyword args, self/cls filtering, ambiguous params, function-local variables, nested functions, variadic params, multi-arg boto3 calls, conflicted function names, LEGB scoping

## Known Limitations

- `boto3.Session().client('s3')` (session-based creation) not tracked - falls back to disambiguation
- Unbound method calls (`Class.method(obj, arg)`) not supported
- Duplicate function names fall back to `None` rather than qualifying by class name


iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
```diff
@@ -0,0 +1,1659 @@
...
+            }
+
+            // Then check parameters (still part of local scope in Python)
+            if let Some(type_infos) = self
+                .parameter_types
+                .get(&(func_name.to_string(), var_name.to_string()))
+            {
+                // Return first match for parameters
+                return type_infos.iter().next().map(|info| &info.service_name);
+            }
```
mschlaipfer: If a parameter has multiple inferred types (e.g., both `s3` and `ec2`). I think we need to conservatively fall back to returning `None` here? Same issue at line 935.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
```diff
@@ -0,0 +1,1659 @@
...
+            for param_node in param_nodes {
+                let param_text = param_node.text().to_string().trim().to_string();
+                // Skip commas and empty strings
+                if param_text == "," || param_text.is_empty() {
+                    continue;
+                }
+                // Remove default values and type annotations
+                if let Some(param_name) = Self::extract_all_params(&param_text).into_iter().next() {
+                    params.push(param_name);
+                }
```
mschlaipfer: **`self`/`cls` not filtered from parameter lists.** If a class method is `def upload(self, client):` and called as `Uploader.upload(uploader, s3)`, positional matching maps `s3` → `self`.

Suggestion: filter `self` and `cls` before pushing to `params`:
```rust
if param_name != "self" && param_name != "cls" {
    params.push(param_name);
}
```

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
```diff
@@ -0,0 +1,1659 @@
...
+                let assign_env = node_match.get_env();
+
+                let new_var = if let Some(node) = assign_env.get_match("NEW") {
+                    node.text().to_string()
+                } else {
+                    continue;
+                };
+
+                let old_var = if let Some(node) = assign_env.get_match("OLD") {
+                    node.text().to_string()
```
mschlaipfer: **Overly broad pattern.** `$NEW = $OLD` matches any assignment (`x = func()`, `x = 42`, etc.). While functionally safe (the HashMap lookup gates progress), this does unnecessary work.

Suggestion: filter by tree-sitter node kind on `$OLD` before the HashMap lookup:
```rust
let old_node = assign_env.get_match("OLD")?;
if old_node.kind() != "identifier" {
    continue;
}
```

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
```diff
@@ -0,0 +1,1659 @@
...
+            if let Some(param_names) = func_params.get(&func_name) {
+                // Match arguments to parameters positionally
+                for (i, arg) in args.iter().enumerate() {
+                    // Get the corresponding parameter name (if it exists)
+                    if let Some(param_name) = param_names.get(i) {
+                        // Check if the argument is a tracked variable (client or resource)
+                        // Clone the type info before the mutable borrow to avoid borrow checker issues
+                        let type_info = self
+                            .get_type_info_for_variable_in_context(arg, None)
+                            .cloned();
```
mschlaipfer: **Keyword arguments produce wrong parameter mapping.** Positional matching fails for `sync_data(dynamodb_client=ddb, s3_client=s3)` — maps `ddb` → `s3_client` (position 0).

Suggestion: either skip arguments containing `=` from positional matching, or parse keyword args and match by name:
```rust
if let Some((key, value)) = arg.split_once('=') {
    // Match by keyword name instead of position
}
```

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
```diff
@@ -0,0 +1,1659 @@
...
+dynamodb = boto3.client('dynamodb')
+sync_data(s3, dynamodb)
+"#;
+        let ast = create_ast(source_code);
+        let mut tracker = VariableTypeTracker::new();
+        tracker.track_boto3_assignments(&ast);
+
+        // Debug: print what we tracked
+        println!("Module scope: {:?}", tracker.module_scope);
+        println!("Parameter types: {:?}", tracker.parameter_types);
```
mschlaipfer: nit: Remove leftover `println!` debugging statements (also at lines 1285, 1293, 1638-1639). The assertion messages are sufficient for test failure diagnostics.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
```diff
@@ -0,0 +1,1659 @@
...
+
+            log::debug!(
+                "Tracked boto3.client assignment at module level: {var_name} -> {service_name}"
+            );
+            self.module_scope.insert(
+                var_name,
+                VariableTypeInfo::from_service_with_kind(service_name, SdkObjectKind::Client),
+            );
+        }
+    }
```
mschlaipfer: **Duplicate logic.** `track_client_assignments` and `track_resource_assignments` are nearly identical — differing only by the pattern string (`boto3.client` vs `boto3.resource`) and `SdkObjectKind`. Consider extracting a shared helper:
```rust
fn track_boto3_factory_assignments(&mut self, root: ..., factory_method: &str, kind: SdkObjectKind) { ... }
```
Same applies to the "is inside function" walk-up pattern repeated at lines 291, 385, 494, 774 — a small `fn is_inside_function(node) -> bool` helper would consolidate those.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
```diff
@@ -0,0 +1,1659 @@
...
+    // ========== Basic Assignment Tracking Tests ==========
+
+    #[test]
+    fn test_track_simple_client_assignment() {
+        let source_code = r#"
+import boto3
+s3_client = boto3.client('s3')
+"#;
+        let ast = create_ast(source_code);
+        let mut tracker = VariableTypeTracker::new();
```
mschlaipfer: **Tests should use `rstest`.** Many tests here (e.g., `test_track_simple_client_assignment`, `test_double_quotes`, `test_client_kind_inference`, `test_resource_kind_inference`) test the same pattern with minor input variations. This crate already uses `rstest` elsewhere (see `service_resolver.rs`). Suggested refactoring:

```rust
use rstest::rstest;

#[rstest]
#[case("boto3.client('s3')", "s3", SdkObjectKind::Client)]
#[case("boto3.client(\"ec2\")", "ec2", SdkObjectKind::Client)]
#[case("boto3.resource('dynamodb')", "dynamodb", SdkObjectKind::Resource)]
fn test_direct_assignment_tracking(
    #[case] rhs: &str,
    #[case] expected_service: &str,
    #[case] expected_kind: SdkObjectKind,
) {
    let source_code = format!("import boto3\nvar = {rhs}\n");
    let ast = create_ast(&source_code);
    let mut tracker = VariableTypeTracker::new();
    tracker.track_boto3_assignments(&ast);

    let info = tracker.get_type_info_for_variable_in_context("var", None).unwrap();
    assert_eq!(info.service_name, expected_service);
    assert_eq!(info.kind, Some(expected_kind));
}
```

Similarly, `test_extract_all_params` at line 1368 would benefit from `#[rstest] #[case(...)]`.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
```diff
@@ -0,0 +1,1659 @@
...
+            .map(|type_infos| {
+                type_infos
+                    .iter()
+                    .map(|info| info.service_name.clone())
+                    .collect()
+            })
+    }
+}
+
+#[cfg(test)]
```
mschlaipfer: **Missing test coverage.** Consider adding tests for:
- Reassignment in same scope: `s3 = boto3.client('s3'); s3 = boto3.client('ec2')` — does second win?
- `async def` functions — does the AST pattern still match?
- Decorated functions: `@retry\ndef upload(client): ...`
- Paginator/waiter skipping (negative test): verify `paginator = client.get_paginator('op')` is NOT tracked
- Keyword args at call site (documents limitation): `upload(client=s3_client)`
- `self` in method definitions

iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
```diff
@@ -82,14 +125,49 @@ impl Extractor for PythonExtractor {
...
         // Find all method calls with attribute access: obj.method(args)
         for node_match in root.find_all(pattern) {
-            if let Some(method_call) = self.parse_method_call(&node_match, source_file) {
-                method_calls.push(method_call);
+            // Determine which function this method call is in (if any)
+            let call_line = node_match.get_node().start_pos().line();
+            let current_function = function_ranges
+                .iter()
+                .find(|(_, range)| range.contains(&call_line))
+                .map(|(name, _)| name.as_str());
```
mschlaipfer, line 177: nit: Consider adding a comment noting the potential optimization:
```rust
// Linear scan is fine for typical Python files. If this becomes a bottleneck,
// sort function_ranges by start line and use binary search (partition_point).
```

iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
```diff
@@ -44,6 +49,44 @@ impl PythonExtractor {
...
+            } else if let Some(func_name) = current_function {
+                // Not found as a variable. Check if it's a parameter with multiple possible types.
+                // get_services_for_parameter returns ALL possible types, not just the first.
+                if let Some(services) = tracker.get_services_for_parameter(func_name, receiver_name)
+                {
+                    let service_vec: Vec<String> = services.iter().cloned().collect();
+                    log::debug!(
+                        "Resolved parameter '{receiver_name}' in function '{func_name}' to multiple services {service_vec:?} for method '{method_name}'"
+                    );
+                    service_vec
```
mschlaipfer, line 86: **Unreachable multi-service branch.** `get_service_for_variable_in_context` already checks `parameter_types` (at variable_type_tracker.rs:890-896), returning the first match. So when it returns `None`, the parameter is NOT in `parameter_types` either — meaning `get_services_for_parameter` at line 66 here will also return `None`. This branch is dead code.

Suggestion: remove the parameter check from `get_service_for_variable_in_context` so the two APIs have distinct responsibilities:
- `get_service_for_variable_in_context` → checks local variables + module scope only
- `get_services_for_parameter` → checks parameters (returns all possible types)

Then this branch becomes reachable and correctly surfaces multiple services to disambiguation.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
mschlaipfer, line 1: Can we create a module and split this into smaller files (more consumable by agents)?

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
adpaco: Right, we should fall back to `None`. Addressed in both places.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
adpaco: Done.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
adpaco: Added, thanks.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
adpaco: Done as part of the restructuring.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
adpaco: These have been merged into `track_boto3_factory_assignments`. Also extracted the `is_inside_function` helper.

iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
adpaco: Done, the branch is now reachable.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
adpaco: We are now parsing keyword args and matching by name.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
adpaco: Done, converted to `rstest` where appropriate.

iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
adpaco: Done.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
adpaco: Makes sense, this file has been split into `mod.rs`, `types.rs`, `tracking.rs`, `lookup.rs`, and `tests.rs`.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker.rs
adpaco: Added.

adpaco:
This is ready for review again. In addition to addressing comments in @mschlaipfer's review, I've made these additional fixes

- **Multi-arg `boto3.client()` calls**: Pattern now uses `$$$ARGS` and extracts the first positional string arg, so `boto3.client('s3', region_name='us-east-1')` is correctly tracked.
- **Duplicate function names**: Methods with the same name in different classes (e.g., `S3Handler.process` and `EC2Handler.process`) are detected and marked conflicted -> lookups return `None` to avoid false narrowing. Note: we could qualify with class name for full resolution, but it's left as future work.
- **Variadic params excluded**: `*args`/`**kwargs` are no longer included in the positional parameter list.
- **Call-site context awareness**: Argument type resolution now uses the calling function's scope, so function-local variables passed to other functions are correctly resolved.
- **Nested function scope in `extractor.rs`**: Changed `.find()` to `.filter().min_by_key(range size)` so method calls inside nested functions resolve to the innermost enclosing function.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker/tracking.rs
```diff
@@ -0,0 +1,658 @@
...
+                    "Tracked boto3.{factory_method} assignment in function '{func_name}': {var_name} -> {service_name}"
+                );
+
+                self.function_scopes
+                    .entry(func_name.clone())
+                    .or_default()
+                    .insert(
+                        var_name,
+                        VariableTypeInfo::from_service_with_kind(service_name, kind.clone()),
+                    );
```
mschlaipfer, line 117: Bug: `func_match.get_node().find_all(...)` searches recursively into nested function definitions. An assignment inside an inner function gets attributed to the outer function's scope as well.

Example that triggers a false positive:
```python
def outer():
    def inner():
        client = boto3.client('s3')
    # outer's scope now incorrectly contains client -> s3
    client.something()  # wrongly resolved to s3
```

`track_function_calls` already handles this correctly using `has_intervening_function`. The same guard is needed here and in `track_aliases` (line 176) and `track_resource_derived_variables` (line 496).

Suggested fix: after matching, skip nodes where `has_intervening_function(&node_match, func_match.get_node().node_id())` returns true.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker/tracking.rs
```diff
@@ -0,0 +1,658 @@
...
+                        type_info.service_name
+                    );
+
+                    self.function_scopes
+                        .entry(func_name.clone())
+                        .or_default()
+                        .insert(new_var, type_info);
+                }
+            }
+        }
```
mschlaipfer: Bug: Function-level aliases are resolved before module-level aliases are processed (module-level starts at line 218). If a function references a module-level alias that hasn't been inserted into `module_scope` yet, the lookup at line 197 fails silently.

Example:
```python
import boto3
s3 = boto3.client('s3')
my_s3 = s3            # module-level alias — not yet in module_scope when func aliases run

def func():
    local = my_s3     # get_type_info_for_variable_in_context("my_s3", Some("func")) -> None
```

Fix: swap the order — process module-level aliases first, then function-level aliases.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker/tracking.rs
```diff
@@ -0,0 +1,658 @@
...
+    }
+
+    /// Track boto3 factory assignments (client or resource) at both function and module level
+    fn track_boto3_factory_assignments(
+        &mut self,
+        root: &ast_grep_core::Node<ast_grep_core::tree_sitter::StrDoc<Python>>,
+        factory_method: &str,
+        kind: SdkObjectKind,
+    ) {
+        let assign_pattern = format!("$VAR = boto3.{factory_method}($$$ARGS)");
```
mschlaipfer, line 75: Limitation: This only matches `boto3.client(...)` / `boto3.resource(...)` directly. A very common pattern in production code is session-based creation:

```python
session = boto3.Session(region_name='us-east-1')
s3 = session.client('s3')
```

I see this is documented in the PR description as a known limitation, but wanted to flag that in my experience this accounts for a significant chunk of real-world boto3 usage (especially in Lambda handlers and multi-region code). Might be worth a follow-up issue to track it.

One approach: first track `$VAR = boto3.Session($$$)` assignments, then also match `$VAR2 = $SESSION.client($$$ARGS)` where `$SESSION` is in the set of known session variables.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker/types.rs
```diff
@@ -0,0 +1,103 @@
...
+#[allow(dead_code)]
+pub(crate) enum SdkObjectKind {
+    /// boto3.client('service') - Low-level client
+    Client,
+    /// boto3.resource('service') - High-level resource
+    Resource,
+    /// client.get_paginator('operation') - Paginator
+    Paginator,
+    /// client.get_waiter('waiter') - Waiter
+    Waiter,
```
mschlaipfer: Nit: `Paginator` and `Waiter` are defined here but never constructed — paginators and waiters are explicitly skipped in `track_resource_derived_variables` (line 517 of tracking.rs). The `#[allow(dead_code)]` on the whole enum suppresses the warning and should be removed.

If these are planned for a future PR, consider adding a TODO comment. Otherwise I'd suggest removing them.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker/tracking.rs
```diff
@@ -0,0 +1,658 @@
...
+
+        for func_match in root.find_all(func_def_pattern) {
+            let env = func_match.get_env();
+            if let Some(node) = env.get_match("FUNC") {
+                let name = node.text().to_string();
+                if !seen.insert(name.clone()) {
+                    self.conflicted_functions.insert(name);
+                }
+            }
+        }
```
mschlaipfer, line 56: The `conflicted_functions` approach is sound as a safety valve, but it's overly conservative for multi-class files (which are common in Python). Two methods named `process` in different classes (`S3Handler.process` and `EC2Handler.process`) would disable tracking for ALL uses of `process`, even though calls like `self.process()` inside each class are unambiguous to a reader.

The TODO mentions qualifying by class name. In the meantime, consider at least limiting the conflict detection to top-level (non-method) functions — i.e., skip function defs that are immediate children of a `class_definition` node. That way, top-level function name collisions are still caught, but class methods don't poison the tracker. The false-narrowing risk from class methods is already low because `self.process()` won't match a bare `process` in `func_params`.

Up to you whether this is worth addressing now vs. in a follow-up.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker/tracking.rs
adpaco: Swapped, thanks!

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker/types.rs
adpaco: Removed.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker/tracking.rs
adpaco: Sounds good - added an `is_method` helper to exclude methods from conflict detection

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker/tracking.rs
adpaco: The PR is getting quite large so I'd rather follow up on a separate one. Opened #226 and linked it here.

iam-policy-autopilot-policy-generation/src/extraction/python/variable_type_tracker/tracking.rs
adpaco: Done.

mschlaipfer (approved):

janstey-amzn (approved):

Issue #58, opened 2025-12-05
Add support for Java language


paulmoloney-visma:
I note that currently this tool initially supports applications built in Python, Go, and TypeScript only . I request that support is added for the JAVA language.

Dianayin422:
Hi @paulmoloney-visma , I'm Diana Yin, the PM for IAM Policy Autopilot. Thanks for sharing your feedback here. I'd love to hear more about your experience. If you're open to a quick chat, feel free to email me at dianayin@amazon.com and we can set up a time. Thanks!

PR #134, opened 2026-02-01
feat: add support for Java language

kdnakt:
*Issue #, if available:*
#58 

*Description of changes:*

Added support for extracting SDK calls from AWS Java SDK v1 & v2.


Dianayin422:
Hi @kdnakt  I'm Diana Yin, the PM for IAM Policy Autopilot. Thanks for sharing your feedback here. 'd love to hear more about your experience. Are there more features or languages that you would like to see in IAM Policy Autopilot? If you're open to a quick chat, feel free to email me at dianayin@amazon.com and we can set up a time. Thanks!

iam-policy-autopilot-policy-generation/src/extraction/java/extractor.rs
```diff
@@ -0,0 +1,198 @@
...
+        let method_name = method_name
+            .strip_suffix("Paginator")
+            .unwrap_or(&method_name)
+            .to_string()
+            .to_case(Case::Pascal);
+
+        Some(crate::SdkMethodCall {
+            name: method_name,
+            possible_services: Vec::new(), // Will be determined later during service validation
+            metadata: None,
```
adpaco: Why do we not store any metadata? We have `_receiver` and `_arguments` available so I don't understand why we don't construct this.

adpaco-aws (commented):
Thank you for the PR, @kdnakt !

This is a solid implementation for adding Java support following the same approach as Javascript for extraction. I do have a minor comment.

mschlaipfer (dismissed):
Unfortunately, I think that in its current shape, this extractor would lead to a large amount of false positive findings. In the JS and TS extractors we use `import` information which we use as a proxy for operations. I could imagine that a Java extractor could work similarly by using `import` information to help with figuring out what "real" operations are. However, it looks like we don't use that right now and just rely on the called method's name.

The problem with just using the method name is that it will lead to operations, which exist in multiple services, being included wrongly as we don't do anything to limit the number of `possible_services`. Say, both service `X` and `Y` have an operation `Foo`, then for a call
```
x_client.foo(...)
```
we would extract `possible_services = [x, y]` because we just match against the service reference, which is essentially a map
```
foo -> [{ service: x, argument_info: ... }, { service: y, argument_info: ... }]
```

How we do this in other extractors is by 1) argument-based disambiguation (for example in Python the arguments need to be passed by argument name and we match them against the API model) and 2) import-based filtering.

I assume that in Java import-based filtering could work quite well.

The extractor would also consider calls that are not SDK calls at all to be considered as SDK calls with higher likelihood. I believe that here the argument-based filtering will help, which can at least check that the number of arguments passed to the method is within the interval `[min_required_args, total_required_and_optional_args]`.

We are going to improe our test setup so these issues become more obvious when running tests.

iam-policy-autopilot-policy-generation/src/extraction/java/extractor.rs
kdnakt: At first I also thought the metadata was needed to extract SDK method calls, but actually it worked without metadata.

But according to [this comment](https://github.com/awslabs/iam-policy-autopilot/pull/134#issuecomment-3835359024), I think I need to construct the metadata to extract SDK method calls for utility functions like S3TransferManager.

I'm working on this now!

kdnakt:
@mschlaipfer san, thank you for your feedback, I'll try use `import` information not to get a lot of false positive findings.

mschlaipfer:
@kdnakt thanks! I'll work on creating a Java test suite so we can see how your extractor performs on it. Happy to answer any questions you may have as well, I am happy to go into more details.

kdnakt:
> Hey @kdnakt, thanks for merging in `main`.
> 
> We're prioritizing Java support, and I'd like to contribute to your PR. Could you enable push permissions to your `java-support` branch? See the "Allow edits from maintainers" part of https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Sorry for being late in updating my PR...

Is this setting enough for you to push commits to my branch?
image

kdnakt:
I have no commits to push for now!

iam-policy-autopilot-policy-generation/resources/config/sdks/boto3
mschlaipfer, line 1: I'll need to check why this has changed. It shouldn't

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -110,8 +110,25 @@ impl Operation {
...
+            //   with no word boundary, so `listObjectsV2` → `ListObjectsV2` (not `ListObjectsV 2`).
+            //   This is the *opposite* of the Python problem, where snake_case splits `V2` into
+            //   `_v_2` and requires a post-processing fix.
+            //
+            // * Mixed-case brand names (WhatsApp, DynamoDB, …): Java SDK v2 preserves the
+            //   internal capitalisation in its camelCase names (e.g. `sendWhatsAppMessage`).
+            //   `convert_case` treats every uppercase letter as a word boundary, so
+            //   `sendWhatsAppMessage` → `SendWhatsAppMessage` — exactly the PascalCase key used
+            //   in the service reference.
+            call.name.to_case(Case::Pascal)
```
mschlaipfer, line 114: My philosophy for the input to enrichment is that they should be method names, not operation names. The idea is to allow for an LLM to do the static analysis, which might not know how a method name maps to an operation.

Currently, it's not how it works (besides keeping the casing and needing map to operations here): Utilities are resolved to operation names during extraction and paginators and waiters are normalized, i.e., stripped `waitUntil..` etc.

I'll address this when working on opening up the enrichment interface, but I think doing this translation here is right (even though eventually it should look different).

iam-policy-autopilot-policy-generation/resources/config/sdks/java-sdk-v2-utilities.json
mschlaipfer, line 1: I have yet to review these myself

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
weibenz1: Not sure why we are want to `The idea is to allow for an LLM to do the static analysis`?

In terms of input, I think it still makes sense to extract operation name then do the enrichment, which aligns with current design of the responsibility separation. Or am I missing something?

mschlaipfer: I thought about this some more, and you are right. We need to have operation names as inputs to enrichment.

My thought was that particularly for utility functions, it would be great if the LLM could use our SDK model to map method names to the underlying operations. However, that doesn't make sense, since we would want to use the LLM for extraction with languages where we don't have a deterministic extraction step yet, hence we also wouldn't have a model of how utility methods map to operations.

iam-policy-autopilot-policy-generation/resources/config/sdks/java-sdk-v2-utilities.json
weibenz1: why do we need this config?

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/import_extractor.rs
```diff
@@ -0,0 +1,210 @@
...
+///
+/// # Panics
+///
+/// Panics on first access if the embedded service-configuration JSON is missing or
+/// malformed — both conditions indicate a corrupt binary and are unrecoverable.
+///
+/// [`SmithyBotocoreServiceNameMapping`]: crate::service_configuration::ServiceConfiguration
+static JAVA_IMPORT_SERVICE_MAP: LazyLock<HashMap<String, String>> = LazyLock::new(|| {
+    load_service_configuration()
+        .expect("service-configuration.json must be present in embedded data")
```
weibenz1, line 33: nit: constant usage?

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/paginator_extractor.rs
```diff
@@ -0,0 +1,151 @@
...
+///       regex: 'Paginator$'
+///       pattern: $PAGINATOR_METHOD
+/// ```
+///
+/// The label `$PAGINATOR_METHOD` is the discriminator (captures the `*Paginator` method name).
+pub(crate) struct JavaPaginatorExtractor;
+
+impl SdkExtractor for JavaPaginatorExtractor {
+    fn rule_yaml(&self) -> &'static str {
+        "kind: method_invocation\nall:\n  - has:\n      field: object\n      pattern: $PAG_CLIENT\n  - has:\n      field: name\n      regex: 'Paginator$'\n      pattern: $PAGINATOR_METHOD"
```
weibenz1: do we want to embed these as config, and/or do you see them evolve overtime? 

might worth introducing a yaml related struct to validate/enforce schema that each extractor uses

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/paginator_extractor.rs
```diff
@@ -0,0 +1,151 @@
...
+            operation,
+            parameters,
+            usage,
+            location,
+            receiver_declaration,
+        });
+    }
+}
+
+/// Detect whether the paginator call result is assigned to a variable.
```
weibenz1: why is this important?

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utility_import_extractor.rs
```diff
@@ -0,0 +1,142 @@
...
+//! [`JavaImportExtractor::process`]: super::import_extractor::JavaImportExtractor
+
+use std::sync::LazyLock;
+
+use crate::extraction::java::extractor::JAVA_UTILITIES_MODEL;
+
+// ── Model-driven import prefix table ─────────────────────────────────────────
+
+/// A single entry in the import-prefix lookup table derived from the model.
+struct ImportEntry {
```
weibenz1, line 32: same concept that I'm implementing in terraform

iam-policy-autopilot-policy-generation/src/extraction/java/matchers/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+/// matchers when the extractor has successfully resolved the receiver variable's
+/// declared type (e.g. `"S3Waiter"`, `"S3AsyncWaiter"`, `"CloudVaultClient"`,
+/// `"CloudVaultAsyncClient"`, or a fully-qualified name).
+///
+/// Resolution order:
+/// 1. **FQN fast path**: if `type_name` starts with `software.amazon.awssdk.services.`,
+///    extract the service name directly from the prefix (no import lookup needed).
+/// 2. **Import lookup**: match `type_name` against the last segment of each specific import
+///    in the file (handles the common `import ...S3AsyncClient;` case).
+/// 3. **Service-id matching**: if no specific import matches (e.g. wildcard import), derive
```
weibenz1, line 49: this is kind of like guessing at this point right?

iam-policy-autopilot-policy-generation/src/extraction/java/extractor.rs
```diff
@@ -0,0 +1,318 @@
...
+pub(crate) type JavaNodeMatch<'a> =
+    ast_grep_core::NodeMatch<'a, ast_grep_core::tree_sitter::StrDoc<Java>>;
+
+// ================================================================================================
+// SdkExtractor trait
+// ================================================================================================
+
+/// Trait implemented by each Java-specific extractor.
+///
+/// Each implementation handles one syntactic pattern (imports, method calls, waiters, etc.).
```
weibenz1, line 80: nice!

iam-policy-autopilot-policy-generation/src/extraction/java/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+    source_files: Vec<SourceFile>,
+    service_index: &ServiceModelIndex,
+) -> Result<Vec<SdkMethodCall>> {
+    if source_files.is_empty() {
+        return Err(ExtractorError::validation("No source files provided"));
+    }
+
+    // Phase 1 — parallel AST extraction (CPU-bound).
+    //
+    // Each file is processed on a blocking thread via `spawn_blocking`.  The
```
weibenz1, line 128: nit: file descriptor issue we ran into before

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/paginator_extractor.rs
mschlaipfer: I don't think it's worth it to come up with a schema and validate it. We would fail at test time, if this was wrong. I think a type that can have a `merge(&self, other: Self)` could make sense, but I'd suggest to do this in a future PR. In this PR I will change the strings to `r#...#` ones so these are more readable.

iam-policy-autopilot-policy-generation/src/extraction/java/mod.rs
mschlaipfer: This is fine according to Claude, am I missing something?
> No file descriptor exhaustion risk. `SourceFile` holds file content as an owned `String` — files are already fully read into memory before `extract_java_sdk_calls` is called. `extract_from_file` operates entirely on `source_file.content` and never opens a file descriptor.

iam-policy-autopilot-policy-generation/resources/config/sdks/java-sdk-v2-utilities.json
mschlaipfer: This is for utility functions which are part of the SDK, which call the AWS APIs, but which themselves are not in the underlying botocore SDK model we use to extract methods by name.

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/paginator_extractor.rs
mschlaipfer: We don't use it, so removing it. Nice catch!

Some context:
I initially added it because in Python (and other extractors) we use the information of how a paginator/waiter is used, but we don't in this extractor--after all. We include an operation when the paginator is created, regardless of whether we detect usage of it.

I ended up not using argument-based disambiguation in Java, and arguments are passed to the paginator constructor (rather than the paginate call in Python), so tracking usage seems not important.

iam-policy-autopilot-policy-generation/src/extraction/java/matchers/mod.rs
mschlaipfer: While there are bugs in this version (fixed in later revisions), I don't think so.

This path means we found the type name of the receiver, it is not fully qualified and there is a wildcard import. (A revised version requires that the wildcard import must be on an SDK import".)

We then see if the type name is an AWS type name and use that information to resolve clashes.

Say
```
import software.amazon.awssdk.services.dynamodb.*;
import software.amazon.awssdk.services.dynamodbv2.*; // made up
```

both provide an operation `ListTables` and in code we have
```
DynamoDbClient client = ...;
client.listTables(...);
```
gets resolved to `dynamodb` as the possible service. Without looking for `DynamoDbClient` (which this code path does), we would not know what service `listTables` is called on.

There was also a bug in how we converted service IDs to type names, fixed since.

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/import_extractor.rs
mschlaipfer: Could you elaborate?

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -227,13 +239,22 @@ This flag has no effect on the generate-policies subcommand."
...
-Optionally takes AWS context (region and account) for accurate ARN generation.\n\n\
-TIP: Use --service-hints to specify the particular AWS services that your application uses if you know them. \
-The final policy may still include actions from other services if required for your operations.")]
+    /// Generates baseline IAM policy documents from source files
+    #[command(
+        long_about = r#"Generates baseline IAM policy documents from source files using
+deterministic static analysis. Optionally takes AWS context (region and account)
+for accurate ARN generation.
+
+Supported languages and SDKs:
```
weibenz1, line 259: down the road, we should have an ENUM or something that defines supported language, and auto generate the supported language section for it.

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/import_extractor.rs
```diff
@@ -0,0 +1,212 @@
...
+/// [`ServiceConfiguration::build_java_import_service_map`] and used by
+/// [`extract_service_from_import`] to normalise extracted service names.
+///
+/// # Panics
+///
+/// Panics on first access if the embedded service-configuration JSON is missing or
+/// malformed — both conditions indicate a corrupt binary and are unrecoverable.
+///
+/// [`SmithyBotocoreServiceNameMapping`]: crate::service_configuration::ServiceConfiguration
+static JAVA_IMPORT_SERVICE_MAP: LazyLock<HashMap<String, String>> = LazyLock::new(|| {
```
weibenz1, line 31: In future, I believe service reference deserves to be in its own crate, and all the service resolution/sdk mapping related logic should be just consolidated there.

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/import_extractor.rs
```diff
@@ -0,0 +1,212 @@
...
+/// the rest of the pipeline.
+///
+/// # Examples
+/// - `"software.amazon.awssdk.services.s3.S3Client"` → `Some("s3")`
+/// - `"software.amazon.awssdk.services.dynamodb.model.GetItemRequest"` → `Some("dynamodb")`
+/// - `"software.amazon.awssdk.services.cloudwatchlogs.CloudWatchLogsClient"` → `Some("logs")`
+/// - `"java.util.List"` → `None`
+fn extract_service_from_import(import_path: &str) -> Option<String> {
+    let segment = import_path
+        .strip_prefix("software.amazon.awssdk.services.")?
```
weibenz1, line 133: does this not trigger the custom lint rule for detecting usage of literal strings rather than defining constants?

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/import_extractor.rs
```diff
@@ -0,0 +1,212 @@
...
+            result.utility_imports.push(UtilityImport {
+                expr: path_part,
+                utility_name,
+                class_name,
+                location,
+            });
+            return;
+        }
+
+        let Some(service) = extract_service_from_import(&path_part) else {
```
weibenz1, line 106: Q: why not do something like `Import::try_from(node, path_parth)`  and `Utility_Import::try_from()`?

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/method_extractor.rs
```diff
@@ -0,0 +1,208 @@
...
+
+        let node = node_match.get_node();
+        let parameters = utils::extract_arguments_from_node(node);
+        let location = Location::from_node(source_file.path.clone(), node);
+        let expr = node.text().to_string();
+
+        // Attempt to find the receiver declaration for simple identifier receivers (Tier 1).
+        // Field-access and method-invocation receivers (Tier 2/3) cannot be resolved without
+        // a type checker, so receiver_declaration is left as None for those cases.
+        let receiver_declaration = match env.get_match("MC_OBJ") {
```
weibenz1: isn't `env.get_match("MC_OBJ")` already populated at line 61?

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/paginator_extractor.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        let node = node_match.get_node();
+        let location = Location::from_node(source_file.path.clone(), node);
+        let expr = node.text().to_string();
+        let parameters = utils::extract_arguments_from_node(node);
+
+        // Attempt to find the receiver declaration for simple identifier receivers (Tier 1).
+        // Field-access and method-invocation receivers (Tier 2/3) cannot be resolved without
+        // a type checker, so receiver_declaration is left as None for those cases.
+        let receiver_declaration = match env.get_match("PAG_CLIENT") {
```
weibenz1: nit: seems to be a refactor opportunity for shared receiver declaration extraction

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
```diff
@@ -0,0 +1,828 @@
...
+    source_file: &SourceFile,
+) -> Option<ReceiverDeclaration> {
+    let call_start = call_node.range().start;
+
+    for ancestor in call_node.ancestors() {
+        let kind = ancestor.kind();
+        let kind_str = kind.as_ref();
+
+        if kind_str == BLOCK {
+            // Scan children of this block that appear before the call node
```
weibenz1: nit: will be more readable and maintainable if we have each dedicated method for each if condition here:

something would look like:
```
switch kind_str:
   case BLOCK: handle_block()
   ... other cases
```

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
```diff
@@ -0,0 +1,828 @@
...
+            for child in ancestor.children() {
+                if child.range().end > call_start {
```
weibenz1: in cases like :

```
class MyClass {
      static client = default();
}
```
and within some static block:
```
static {
   client.methodCall();
}
````
I guess we will not be able to capture it?

File: iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs [hunk 18]
```diff
@@ -0,0 +1,828 @@
...
+            // Note: there is NO intermediate `type_pattern` wrapper in this grammar version.
+            // The binding variable is the last `identifier` child of `instanceof_expression`,
+            // and the declared type is the `type_identifier` child immediately before it.
+            //
+            // No position guard is needed: the call is inside the then-block, which is
+            // always textually after the condition, so the binding is always in scope.
+            for child in ancestor.children() {
+                if child.kind().as_ref() == PARENTHESIZED_EXPRESSION {
+                    for paren_child in child.children() {
+                        if paren_child.kind().as_ref() == INSTANCEOF_EXPRESSION {
```
weibenz1: do we need to constraint the extraction to the same `if instanceof` branch? like escaping all the `else` branchs

iam-policy-autopilot-policy-generation/src/extraction/java/matchers/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Returns an empty list if no candidates match the imports; callers must not emit
+/// an [`SdkMethodCall`] in that case.
+///
+/// [`SdkMethodCall`]: crate::extraction::SdkMethodCall
+pub(crate) fn apply_import_filter(
+    services: Vec<String>,
+    imported_services: &HashSet<String>,
+) -> Vec<String> {
+    if imported_services.is_empty() {
+        return services;
```
weibenz1: this should be an error right

iam-policy-autopilot-policy-generation/src/extraction/java/matchers/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+    // Slow path: simple (unqualified) type name — look up service via the file's import list.
+    // The import path's last segment (e.g. "S3AsyncClient") is matched against `type_name`.
+    let service = file_imports
+        .iter()
+        .find(|imp| imp.expr.split('.').next_back() == Some(type_name))
+        .map(|imp| imp.service.clone());
+
+    if let Some(svc) = service {
+        // Filter the full candidate list to only the pinned service
+        all_services.into_iter().filter(|s| s == &svc).collect()
```
weibenz1, line 85: does it have to be a vector instead of a set?

iam-policy-autopilot-policy-generation/src/extraction/java/extractor.rs
```diff
@@ -0,0 +1,318 @@
...
+
+impl JavaLanguageExtractorSet {
+    /// Build the default extractor set for AWS SDK for Java v2.
+    ///
+    /// Includes extractors for:
+    /// - Import declarations
+    /// - Direct method calls (`receiver.method(args)`)
+    /// - Waiter usages (init and call)
+    /// - Paginator usages
+    pub(crate) fn default_aws_v2() -> Self {
```
weibenz1, line 161: Something I came across before so asked Claude to write a sample code on inventory:
```
// In your trait definition crate
use inventory;

pub(crate) trait Extractor: Send + Sync + 'static {
    fn extract(&self, node: &Node) -> Option<Finding>;
    fn priority(&self) -> u32 { 50 }
}

// A wrapper since inventory needs a concrete type
pub(crate) struct ExtractorEntry {
    pub constructor: fn() -> Box<dyn Extractor>,
}

inventory::collect!(ExtractorEntry);

// In each extractor file — fully decentralized
inventory::submit! {
    ExtractorEntry { constructor: || Box::new(JavaPaginatorExtractor) }
}

// construction — never needs editing again
pub(crate) fn default_aws_v2() -> Self {
    let mut extractors: Vec<Box<dyn Extractor>> = inventory::iter::<ExtractorEntry>
        .map(|entry| (entry.constructor)())
        .collect();
    extractors.sort_by_key(|e| e.priority());
    Self { extractors }
}
```

iam-policy-autopilot-policy-generation/src/extraction/java/matcher.rs
```diff
@@ -0,0 +1,144 @@
...
+impl<'a> JavaMatcher<'a> {
+    /// Create a new matcher backed by the given service index and utilities model.
+    ///
+    /// The caller is responsible for supplying the model (typically the process-wide
+    /// [`JAVA_UTILITIES_MODEL`] static) so the embedded JSON is parsed only once.
+    ///
+    /// [`JAVA_UTILITIES_MODEL`]: crate::extraction::java::extractor::JAVA_UTILITIES_MODEL
+    pub(crate) fn new(
+        service_index: &'a ServiceModelIndex,
+        utility_model: &'a JavaUtilitiesModel,
```
weibenz1, line 46: should the metadata passed in be baked into the constructor, rather than asking the caller to pass it?

iam-policy-autopilot-policy-generation/src/lib.rs
```diff
@@ -60,22 +60,44 @@ pub enum Language {
...
     }
+
+    /// Returns all supported languages.
+    ///
+    /// This is the canonical list of languages that the policy generation crate supports.
+    /// Tests that cover all languages (e.g. the waiter extraction integration test) should
+    /// iterate over this slice so that adding a new language automatically causes those tests
+    /// to fail until the new language is wired up.
+    #[must_use]
+    pub fn supported() -> &'static [Self] {
```
weibenz1, line 84: in our future doc gen work, this should replace hardcoded language support in the cli description

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/import_extractor.rs
afilieri-aws, line 133: Should we include also the prefix `software.amazon.awssdk.enhanced`?

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
```diff
@@ -0,0 +1,828 @@
...
+                                        return Some(decl);
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+
+            // Stop at the class/record boundary — do not walk into enclosing types.
```
afilieri-aws: This is in general a reasonable idea, including for handling inner classes' fields that shadow outer classes'. However, it may not work in case of anonymous classes that can refer to objects in their enclosing scopes. For example:
```
class MyService {
    S3Client s3 = S3Client.create();

    void doWork() {
        S3Client localS3 = S3Client.builder()...build();

        executor.submit(new Runnable() {
            @Override
            public void run() {
                s3.putObject(...);        // refers to MyService.this.s3
                localS3.putObject(...); // refers to the local variable in the method
            }
        });
    }
}
```
That said, the current approach seems a reasonable trade-off to start with. If we observe measurable impact on precision (we may not be able to link the method call to the correct receiver, leading to overapproximation), we can revisit it later.

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
```diff
@@ -0,0 +1,828 @@
...
+/// which only needs `type_name`.
+fn extract_declaration_from_local_var_decl(
+    decl_node: &ast_grep_core::Node<ast_grep_core::tree_sitter::StrDoc<ast_grep_language::Java>>,
+    var_name: &str,
+    source_file: &SourceFile,
+) -> Option<ReceiverDeclaration> {
+    let mut children = decl_node.children();
+
+    // First child is the declared type
+    let type_node = children.next()?;
```
afilieri-aws, line 552: Annotations, modifiers, and their combinations could be corner cases to consider here. For example:
```
final S3Client client = ...;
@Annotated S3Client client = ...;
@Annotated final S3Client client = ...;
final @Annotated S3Client client = ...;

// or with var:
final var client = ...;
@Annotated var client = ...;
@Annotated final var client = ...;
final var @Annotated client = ...;
```

While the extractor would fall back to imports extractions in these cases (should not lose recall), we can probably be more precise with little effort here.

afilieri-aws: I played a bit with it. You can confirm the issue with a test like:

```
class FinalLocalVariable {
    void run() {
        final CloudVaultClient client = CloudVaultClient.create();
        client.describeResource(request);
    }
}

```

and it can be fixed as done for fields in this same file (TreeSitter builds a similar structure):
```
    if children
        .peek()
        .map(|n| n.kind().as_ref() == MODIFIERS)
        .unwrap_or(false)
    {
        children.next();
    }
```

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
```diff
@@ -0,0 +1,828 @@
...
+    let children: Vec<_> = resource_node.children().collect();
+
+    // Need at least: <type>, identifier, "=", <initializer>
+    if children.len() < 2 {
+        return None;
+    }
+
+    // First child is the declared type (may be preceded by modifiers in some grammars,
+    // but the standard resource node starts directly with the type).
+    let type_node = &children[0];
```
afilieri-aws: Similar issue with modifiers here, e.g.:
```
try (final S3Client s3 = S3Client.create()) {
    s3.putObject(request);
}
```

Something like the following should helo:
```
    let offset = if children[0].kind().as_ref() == MODIFIERS { 1 } else { 0 };

    let type_node = &children[offset];
```

(Notice that here the issue is a little worse because the code in the example would reach ln 594 and return `None`, which would extend the search to outer scopes and may match the wrong object outside)

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
```diff
@@ -0,0 +1,828 @@
...
+    param_node: &ast_grep_core::Node<ast_grep_core::tree_sitter::StrDoc<ast_grep_language::Java>>,
+    var_name: &str,
+    source_file: &SourceFile,
+) -> Option<ReceiverDeclaration> {
+    let children: Vec<_> = param_node.children().collect();
+    if children.len() < 2 {
+        return None;
+    }
+
+    let type_text = children[0].text().to_string();
```
afilieri-aws: Same issue with modifiers (and potential fix) as for `extract_declaration_from_resource`

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
```diff
@@ -0,0 +1,828 @@
...
+/// - `string_literal`                  → strip surrounding `"…"` quotes
+/// - `decimal_integer_literal`         → keep text as-is
+/// - `decimal_floating_point_literal`  → keep text as-is
+/// - `hex_integer_literal`             → keep text as-is
+/// - `octal_integer_literal`           → keep text as-is
+/// - `binary_integer_literal`          → keep text as-is
+/// - `true` / `false`                  → keep text as-is
+/// - `null_literal`                    → keep text as-is
+/// - `character_literal`               → strip surrounding `'…'` quotes
+pub(crate) fn resolve_java_literal(
```
afilieri-aws, line 922: Not critical because they are anyway caught decently in the default case, but `hex_floating_point_literal` and `text_block` may be added (the former is probably more relevant for this application than the latter)

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
```diff
@@ -0,0 +1,828 @@
...
+    }
+}
+
+/// Extract positional arguments from a `method_invocation` AST node.
+pub(crate) fn extract_arguments_from_node(
+    node: &ast_grep_core::Node<ast_grep_core::tree_sitter::StrDoc<ast_grep_language::Java>>,
+) -> Vec<Parameter> {
+    let mut parameters = Vec::new();
+
+    let arg_list = node
```
afilieri-aws, line 966: There is a corner case here when there are comments in the arguments list, e.g.,
```
  s3Client.putObject(
      request,  // a comment
      body      // another comment
  );
```
the comments would be extracted as arguments themselves. As a simple solution, you may add comments to the filtering at line 815, i.e., skipping also:
```
const LINE_COMMENT: &str = "line_comment";
const BLOCK_COMMENT: &str = "block_comment";
```

iam-policy-autopilot-cli/src/main.rs
mschlaipfer: Good point, we already have such an enum: https://github.com/mschlaipfer/iam-policy-autopilot/blob/5073c460b529e38e42fdd5f177bf62edd8c2ac00/iam-policy-autopilot-policy-generation/src/lib.rs#L58. I'll create an issue

mschlaipfer: I'm not sure if we should use the enum to generate this, or have a more general template-based approach, opened https://github.com/awslabs/iam-policy-autopilot/issues/169

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/import_extractor.rs
mschlaipfer: Created https://github.com/awslabs/iam-policy-autopilot/issues/170

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/import_extractor.rs
mschlaipfer: No, that is only checking `==` expressions.

mschlaipfer: No, utility imports are handled separately in `utility_import_extractor`, but it turns out we didn't correctly handle wildcard imports there, fixing that.

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/method_extractor.rs
mschlaipfer: I did the refactoring suggested in your comment in `paginator_extractor.rs`.

As part of it I've also decided to remove `receiver` altogether. It was just used
1) to populate `SdkMethodCallMetadata`, but it's unused there (other languages populate it, but I do not see the point--we could add it back if needed later)
1) in a fallback in the utility matcher, but that didn't make much sense because I looked for a class name, which we do not necessarily store here as part of `receiver` (which is an arbitrary expression).

That refactor also removed fields from the expected results of the test fixtures, so unfortunately that single commit is large, but generally makes sense IMHO.

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/paginator_extractor.rs
mschlaipfer: I will do this this refactoring as part of this PR

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
mschlaipfer: These work already, adding tests

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
mschlaipfer: This should be okay because of the name guard in `extract_declaration_from_instanceof_expr`

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
mschlaipfer: Nice catch, added support.

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
mschlaipfer: Great catch, fixed

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
mschlaipfer: `text_block` doesn't exist in the tree-sitter grammar, but I've added handling of triple-quoted strings and `hex_floating_point_literal`

iam-policy-autopilot-policy-generation/src/extraction/java/extractors/utils.rs
mschlaipfer: Fixed, nice catch! Thanks!

iam-policy-autopilot-policy-generation/src/extraction/java/matchers/mod.rs
mschlaipfer: Nice catch! It should be removed to go from possible false positive to possible false negative (in case we miss something during import extraction).

Also fixing test fixtures where I did not add import statements before.

iam-policy-autopilot-policy-generation/src/extraction/java/matchers/mod.rs
mschlaipfer: Let's revisit if needed.

iam-policy-autopilot-policy-generation/src/extraction/java/extractor.rs
mschlaipfer: Thanks, wasn't aware of this. I assume you're referring to https://crates.io/crates/inventory? I have a followup PR to factor the Java extractor out into a framework. I'll take this into account then.

iam-policy-autopilot-policy-generation/src/extraction/java/matcher.rs
mschlaipfer: Do you mean `result: &ExtractionResult` should be passed here? Since I refactor this into a framework, I'll leave this as is now, but I think you're right, it could be passed to the constructor. We might not need a `JavaMatcher` type at all.

iam-policy-autopilot-policy-generation/src/lib.rs
mschlaipfer: also https://github.com/awslabs/iam-policy-autopilot/issues/169

weibenz1 (dismissed):

afilieri (approved):

afilieri-aws (dismissed):

weibenz1 (dismissed):

adpaco-aws (dismissed):

weibenz1 (approved):

adpaco-aws (approved):

afilieri-aws (approved):

Issue #132, opened 2026-01-30
[CI] PR checks should validate version format


weibenz1:
PR checks should have a step to validate the version format of iam-policy-autopilot binaries

PR #137, opened 2026-02-03
feat: automated submodule update

weibenz1:
*Issue #, if available:*
https://github.com/awslabs/iam-policy-autopilot/issues/63

*Description of changes:*
This PR implements the automated git submodule update. At high level, the workflow runs weekly on Sunday that does:
1. Download latest `iam-policy-autopilot` from pypi and get compiled boto3/botocore datahash
2. Pull repository and compile from source, and get the built boto3/botocore datahash
3. compare pypi against built datahash
4. If there's a diff, create a new branch and release it.

Note: The auto update only checkout latest tags from both submodules, rather than just latest commit in the main branch - this is how we ensure it's latest released change.

Testing & Validation:
1. Pull request opened by bot https://github.com/weibenz1/iam-policy-autopilot/pull/5
2. The actual build that resulted in the pull request - https://github.com/weibenz1/iam-policy-autopilot/actions/runs/21652828546

### DIFF
--- a/.github/workflows/weekly_submodule_update.yml
+++ b/.github/workflows/weekly_submodule_update.yml
@@ -1,0 +1,142 @@
+name: Weekly Submodule Update Check
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'  # Sunday at midnight UTC
+  workflow_dispatch:
+
+jobs:
+  check_versions:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      has_diff: ${{ steps.compare.outputs.has_diff }}
+      pypi_output: ${{ steps.pypi.outputs.output }}
+      built_output: ${{ steps.built.outputs.output }}
+      boto3_tag: ${{ steps.submodules.outputs.boto3_tag }}
+      botocore_tag: ${{ steps.submodules.outputs.botocore_tag }}
+    steps:
+    - name: Get PyPI version output
+      id: pypi
+      run: |
+        pip install iam-policy-autopilot
+        OUTPUT=$(iam-policy-autopilot --version --verbose)
+        echo "output<<EOF" >> $GITHUB_OUTPUT
+        echo "$OUTPUT" >> $GITHUB_OUTPUT
+        echo "EOF" >> $GITHUB_OUTPUT
+        echo "PyPI version output:"
+        echo "$OUTPUT"
+
+    - uses: actions/checkout@v4
+      with:
+        submodules: recursive
+        fetch-depth: 0
+
+    - name: Update submodules to latest tags
+      id: submodules
+      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/weekly_submodule_update.yml
```diff
@@ -0,0 +1,137 @@
...
+        echo "output<<EOF" >> $GITHUB_OUTPUT
+        echo "$OUTPUT" >> $GITHUB_OUTPUT
+        echo "EOF" >> $GITHUB_OUTPUT
+        echo "Built version output:"
+        echo "$OUTPUT"
+
+    - name: Compare versions
+      id: compare
+      run: |
+        extract_hash() { echo "$1" | grep "$2" | sed 's/.*data_hash=\([^ ]*\).*/\1/'; }
```
mschlaipfer, line 72: Can this fail, say if we break our version output format, which might lead to us missing an update because we get the same error code in all variables we compare?

Do we need to have an additional check that our output format is correct?

.github/workflows/weekly_submodule_update.yml
```diff
@@ -0,0 +1,137 @@
...
+        submodules: recursive
+        fetch-depth: 0
+
+    - name: Update submodules and create PR
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        cd iam-policy-autopilot-policy-generation/resources/config/sdks/boto3
+        git fetch --tags
+        git checkout $(git describe --tags $(git rev-list --tags --max-count=1))
```
mschlaipfer: Could we reuse the tag we did the comparison with rather than getting it again?

mschlaipfer:
neat

.github/workflows/weekly_submodule_update.yml
weibenz1: Yes, if we break the version output format, this action would, rightfully, fail

We have issue #132 for this

.github/workflows/weekly_submodule_update.yml
weibenz1: yeah, that's more stable, will do

weibenz1: updated and new build - https://github.com/weibenz1/iam-policy-autopilot/actions/runs/21682229820/job/62519617277

as well as new pull request - https://github.com/weibenz1/iam-policy-autopilot/pull/6

.github/workflows/weekly_submodule_update.yml
adpaco: Can we at least validate extraction on one of the hashes? For example:
```
        if [[ -z "$PYPI_BOTO3_HASH" ]]; then echo "Failed to extract hash"; exit 1; fi
```
We shouldn't be comparing hashes if they're all empty.

PR #139, opened 2026-02-04
refactor: replace duplicated `CallInfo` with shared `WaiterCallPattern`

### LINKED ISSUE #88: Refactor: Call information should be stored in a single type
We currently store call information in various types, each with small specialization, e.g., `ChainedWaiterCallInfo`, `WaitCallInfo`, etc. See attached screenshot (which might not be complete):

Image

We should refactor this so we avoid code duplication. We could have common and a specialized fields, and we should implement getters on these.

Having separate types led to a lot of duplicated work in https://github.com/awslabs/iam-policy-autopilot/pull/84.
https://github.com/awslabs/iam-policy-autopilot/pull/79 works around it (for Python waiters) by introducing a `CallInfo` enum, but this should be removed as part of this refactoring.

adpaco-aws:
*Issue #, if available:* Closes #88

*Description of changes:*Replace the duplicated `CallInfo` enum in Python and Go waiter extractors
with a shared `WaiterCallPattern` enum in `extraction_utils.rs`. This seemed like the best option to me considering that the alternative was to either
 1. replicate the logic through synthetic call creation methods for each variant (unmatched, matched and chained)
 2. follow a trait-based approach which feels over-complicated for this use case

In fact, we're currently doing (1) for paginator extraction leading to a significant amount of code duplication. If we agree on keeping an enum to drive this logic, I can follow up with a refactor to use `WaiterCallPattern` (with a better name) on the paginator extraction code.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/enrichment/operation_fas_map.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/operation_fas_map.rs
@@ -98,7 +98,11 @@ where
                         .into_iter()
                         .filter_map(|v| v.as_str().map(|s| s.to_string()))
                         .collect(),
-                    _ => continue, // Skip non-string/non-array values
+                    _ => {
+                        return Err(serde::de::Error::custom(
+                            format!("Expected String or String array in FAS context values for key '{}', got: {:?}", key, value)
+                        ));
+                    }
                 };
 
                 if !values.is_empty() {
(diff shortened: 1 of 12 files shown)


iam-policy-autopilot-policy-generation/src/extraction/shared/extraction_utils.rs
```diff
@@ -89,3 +89,94 @@ pub(crate) struct ChainedPaginatorCallInfo {
...
+    },
+
+    /// Chained waiter call (Python-specific)
+    ///
+    /// Used for `client.get_waiter('name').wait(args)` pattern.
+    /// Synthetic calls use arguments from the chained call.
+    ///
+    /// Note: Go SDK does not support this pattern, so Go extractors
+    /// will never construct this variant.
+    Chained(&'a ChainedWaiterCallInfo),
```
mschlaipfer, line 126: I've been wondering if we couldn't just use `Matched` for this, where we split the chain into creation and wait call.

iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
```diff
@@ -293,39 +262,30 @@ impl<'a> WaitersExtractor<'a> {
...
-    ) -> Vec<SdkMethodCall> {
+    /// Create synthetic SdkMethodCalls from a waiter call pattern
+    ///
+    /// This is the unified method for creating synthetic calls from any waiter pattern:
+    /// - CreationOnly: Uses required parameters from service model
+    /// - Matched: Uses arguments from the wait() call
+    /// - Chained: Uses arguments from the chained call
+    ///
+    /// Creates one call per candidate service with the actual operation name.
+    fn create_synthetic_calls_internal(&self, pattern: WaiterCallPattern) -> Vec<SdkMethodCall> {
```
mschlaipfer: Could this be a method on `WaiterCallPattern` that we share between Python and Go?

iam-policy-autopilot-policy-generation/src/extraction/shared/extraction_utils.rs
adpaco: I also noticed this in the previous refactoring but decided to keep it because it represents the actual code structure better. It can be done, but it forces us to keep track of inaccurate expression and locations, requiring a special case in `match_wait_to_waiter` and essentially hiding the fact that this is a valid pattern in Python. Please let me know what you think.

iam-policy-autopilot-policy-generation/src/extraction/python/waiters_extractor.rs
adpaco: Yes! Moved the method to `WaiterCallPattern` in the last commit.

iam-policy-autopilot-policy-generation/src/extraction/shared/extraction_utils.rs
mschlaipfer: Okay, if we cannot split the chained expression and it leads to inaccuracies, probably best to keep it.

PR #145, opened 2026-02-09
feat(mcp): enable policy explanations in generate_policy tool

Oluwatobi-Mustapha:
### Context
The core `iam-policy-autopilot` engine can generate reasoning data (mapping permissions to source code lines), but the MCP server was previously discarding this data. This limited AI agents’ ability to understand *why* a specific permission was granted.

### Changes
- **Schema extension:** Updated `GeneratePoliciesInput` to accept an optional `explain` boolean, and `GeneratePoliciesOutput` to return a serialized `explanations` string.
- **Data plumbing:** Updated `generate_application_policies` logic to pass the explanation filter to the core engine and serialize the results.
- **Core visibility:** Promoted the `enrichment` module in `iam-policy-autopilot-policy-generation` from `pub(crate)` to `pub` to enable robust integration testing of explanation data structures.
- **Documentation:** Added missing documentation to enrichment structs/enums to satisfy crate-level lint requirements (`#![deny(missing_docs)]`).
- **Testing:** Added a new integration test `test_generate_application_policies_with_explanations` to verify end-to-end data flow.

### Visual proof

**Before (Gap Identified):** ***Output contains only the policy, lacking context.*** 
C5E81391-761B-4B33-8E8B-CD2892D03C62_1_201_a

**After:** ***Output now includes detailed reasoning mapping actions (e.g., `s3:ListBucket`) to source locations.***
Image 2026-02-09 at 1 23 AM

### Testing

Ran:
```bash
cargo test --package iam-policy-autopilot-mcp-server
```

✅ Result: All tests passed, including the new `test_generate_application_policies_with_explanations`.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
@@ -59,6 +59,7 @@ impl Operation {
     ///
     /// This is public to allow downstream integration tests (like in the MCP server)
     /// to construct realistic mock data.
+    #[must_use]
     pub fn new(service: String, name: String, source: OperationSource) -> Self {
         Self {
             service,
@@ -216,6 +217,7 @@ impl Explanations {
     const FAS: &str =
         "The explanation contains an operation added due to Forward Access Sessions (FAS). See https://docs.aws.amazon.com/IAM/latest/UserGuide/access_forward_access_sessions.html.";
 
+    #[must_use]
     pub(crate) fn new(explanations: BTreeMap<String, Explanation>) -> Self {
         let mut documentation: Vec<&'static str> = vec![];
         for explanation in explanations.values() {
--- a/iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
@@ -269,6 +269,8 @@ where
 #[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
 pub(crate) struct ServiceReferenceMapping {
     // represents the top level service reference mapping
+    // FIX: Using custom deserializer to handle reqwest::Url
+    #[serde(deserialize_with = "deserialize_url_map")]
     pub(crate) service_reference_mapping: HashMap<String, Url>,
 }
 
@@ -410,9 +412,6 @@ impl RemoteServiceReferenceLoader {
     }
 
     fn get_cache_dir() -> PathBuf {
-        // not using tempfile crate
-        // instead, using the std to resolve temp dir and then manage the file itself
-        // file deletion is delegated to the OS.
         let cache_dir = std::env::temp_dir().join(IAM_POLICY_AUTOPILOT);
         let _ = std::fs::create_dir_all(&cache_dir);
         cache_dir
@@ -445,7 +444,6 @@ impl RemoteServiceReferenceLoader {
             }
         }
 
-        // check temp file
         let cache_path = Self::get_cache_path(service_name);
         if !self.disable_file_system_cache && Self::is_cache_valid(&cache_path).await {
             if let Ok(content) = fs::read_to_string(&cache_path).await {
@@ -497,7 +495,6 @@ impl RemoteServiceReferenceLoader {
                             e,
                         )
                     })?;
-                // persist content into the temp file as well
                 if !self.disable_file_system_cache {
                     let _ = fs::write(&cache_path, &service_reference_content).await;
                 }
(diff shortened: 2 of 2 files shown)


iam-policy-autopilot-policy-generation/src/lib.rs
```diff
@@ -12,7 +12,7 @@
 pub(crate) mod errors;
 
 // Re-export the enrichment module for public use
-pub(crate) mod enrichment;
+pub mod enrichment;
```
mschlaipfer: I am curious why this needs to be made `pub`. The CLI seems to not need that. I'll need to take a look how the CLI consumes this. Maybe it uses the paths from the `pub use` on line 37 of this file?

Although I see this pattern also elsewhere in this file, visibility there (and more generally) should also be reduced. From my experience, coding agents are prone to make things `pub` more than needed, this might be a remnant of initial development. I also see types like `Operator` which you had to add `///` comments to so they pass the linter, which I assume don't need to be `pub`.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -41,6 +41,7 @@ pub struct Reason {
     pub operations: Vec<Arc<Operation>>,
 }
 
+/// Represents a single IAM operation (Service + Action)
```
mschlaipfer: This comment isn't quite accurate. An operation name is not the action. It should also mention the operation source.

iam-policy-autopilot-mcp-server/src/tools/generate_policy.rs
```diff
@@ -244,4 +278,56 @@ mod tests {
...
+
+        // 6. VERIFY
+        assert!(result.is_ok());
+        let output = result.unwrap();
+
+        assert!(output.explanations.is_some());
+        let explanation_str = output.explanations.unwrap();
+
+        println!("Final Output: {}", explanation_str);
+        assert!(explanation_str.contains("s3:ListBucket"));
```
mschlaipfer, line 356: Since this essentially tests JSON serialization, we should provide more realistic mock data, which populates the reasons and documentation as well.

Eventually (not for this PR) we should refactor so we don't need to mock the return value here.

iam-policy-autopilot-mcp-server/src/tools/generate_policy.rs
```diff
@@ -30,6 +30,11 @@ pub struct GeneratePoliciesInput {
         description = "List of AWS service names to filter SDK calls by (e.g., ['s3', 'dynamodb']). When provided, the result of source code analysis will be restricted to the provided services. The generated policy may still contain actions from a service not provided as a hint, if IAM Policy Autopilot determines that the action may be needed for the SDK call."
     )]
     pub service_hints: Option<Vec<String>>,
+
+    #[schemars(
+        description = "If set to true, the tool will return detailed explanations for why each permission was added. Defaults to false."
```
mschlaipfer, line 35: From your experience, how does the agent populate this typically?

mschlaipfer (changes requested):

iam-policy-autopilot-policy-generation/src/lib.rs
Oluwatobi-Mustapha: Agreed. I've reverted the module visibility to `pub(crate)` and am now strictly re-exporting only the types required for the test (Explanations, Reason, etc). Fixed in my latest commit `278dce5`

iam-policy-autopilot-mcp-server/src/tools/generate_policy.rs
Oluwatobi-Mustapha: In my experience with MCP clients (Claude/Zed), the agent relies on the schema description. Since we explicitly describe the flag (If set to true...), the agent correctly interprets the user's intent when they ask for 'reasoning' or 'why'.

iam-policy-autopilot-mcp-server/src/tools/generate_policy.rs
Oluwatobi-Mustapha: Good point. I've updated the test to construct a realistic `Explanation` object with a concrete Reason and `OperationSource::Provided`. This now validates that the serialization logic works for nested structures. Also fixed in my latest commit.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
Oluwatobi-Mustapha: Thank you! I've updated the doc comment to accurately reflect that the struct contains the Service, Name, and Source. Fixed in my latest commit `87608fd`

karanjitsingh:
Hi @Oluwatobi-Mustapha , hanks for the contribution, I'd like to add one small thing here, I think we need to add a specific instruction in the [tool prompt](https://github.com/awslabs/iam-policy-autopilot/blob/main/iam-policy-autopilot-mcp-server/src/mcp.rs#L112) on the lines of "When the user asks for explanation use the `explain` flag for a detailed output.".

PR #149, opened 2026-02-11
refactor: replace duplicated paginator methods with shared `PaginatorCallPattern`

adpaco-aws:
Issue #, if available: Follow-up to #139

Description of changes:

Replace the duplicated synthetic call creation methods in Python and Go paginator extractors with a shared `PaginatorCallPattern` enum in `extraction_utils.rs`. This follows the same pattern established in PR #139 for waiter extraction.

This eliminates 5 methods related to synthetic call creation (~170 lines of duplicated code) and makes the paginator extraction logic consistent with the waiter extraction approach.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/extraction/shared/extraction_utils.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/shared/extraction_utils.rs
@@ -184,26 +184,25 @@ impl<'a> PaginatorCallPattern<'a> {
     /// Create synthetic SDK method call for this paginator pattern
     ///
     /// This method encapsulates the common logic for creating synthetic calls across
-    /// extractors for different languages. Language-specific behavior is provided via callbacks.
+    /// extractors for different languages.
     ///
     /// # Arguments
     /// * `service_index` - Service model index for method lookup
-    /// * `operation_to_method` - Callback to convert operation name to method name (language-specific)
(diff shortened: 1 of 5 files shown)


iam-policy-autopilot-policy-generation/src/extraction/shared/extraction_utils.rs
```diff
@@ -90,6 +90,146 @@ pub(crate) struct ChainedPaginatorCallInfo {
...
+    /// # Arguments
+    /// * `service_index` - Service model index for method lookup
+    /// * `operation_to_method` - Callback to convert operation name to method name (language-specific)
+    ///
+    /// # Returns
+    /// Synthetic SDK method call with all services that provide this operation
+    pub(crate) fn create_synthetic_call<F>(
+        &self,
+        service_index: &crate::extraction::sdk_model::ServiceModelIndex,
+        operation_to_method: F,
```
mschlaipfer: I think just passing in the language here, and calling the method in `ServiceDiscovery` would be nicer. We already have a utility method for renaming, I don't think the language-specific extractors should ever need to overwrite it.

adpaco: That's indeed nicer. Just updated the code to pass the language to `create_synthetic_calls` (also in the shared waiters code).

PR #150, opened 2026-02-13
feat: add LSP client for Python type information extraction

adpaco:
*Issue #, if available:* N/A

*Description of changes:*

Implements a generic async LSP client built on [`async-lsp`](https://crates.io/crates/async-lsp) for type information extraction from Python code. The client communicates with [the ty Python type checker](https://docs.astral.sh/ty/) and is designed to support additional language servers (e.g., gopls for Go) via the `LspServerConfig` trait.

### Changes
- Add generic `LspClient<C: LspServerConfig>` with `TyConfig` implementation for ty
- Use `async-lsp` for LSP protocol handling (replacing hand-rolled JSON-RPC framing from the initial PR)
- Add per-URI diagnostics tracking (`publishDiagnostics`) so `open_document` returns as soon as the server is ready, with configurable timeout fallback
- Add `LspClientOptions` for tuning timeouts (10s initialize, 5s hover, 2s shutdown, 1s open-document)
- Add `extract_type_from_hover` as a standalone public function with rstest-parameterized unit tests
- Add 7 integration tests with `#[ignore]` + `#[serial]`, covering hover, multi-document, empty file, invalid position, idempotent open, and server-not-found
- Add dedicated CI job that installs ty + boto3-stubs and runs the ignored tests

### Notes
- LSP client is standalone and not yet integrated with the extraction pipeline
- Integration tests require ty and boto3-stubs. They run in CI automatically, or locally with:
```
pip install 'ty>=0.0.39,<0.1' 'boto3-stubs[essential]'
cargo test -p iam-policy-autopilot-policy-generation --features integ-test -- --ignored
```

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -186,6 +186,32 @@ jobs:
         with:
           args: "format --check ."
 
+  lsp-integration:
+    name: LSP Integration Tests
+    needs: [changes]
+    if: needs.changes.outputs.rust == 'true'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Install LSP test dependencies
+        run: |
+          python3 -m venv .venv
+          source .venv/bin/activate
+          pip install 'ty>=0.0.39,<0.1' 'boto3-stubs[essential]'
+          echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
+
+      - name: Run LSP integration tests
+        run: cargo test -p iam-policy-autopilot-policy-generation --features integ-test -- --ignored
+
   unused-deps:
     name: Unused Dependencies (cargo-machete)
     needs: [changes]
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -203,7 +203,10 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Install LSP test dependencies
-        run: pip install --break-system-packages 'ty>=0.0.39,<0.1' boto3 'boto3-stubs[essential]'
+        run: |
+          python3 -m venv .lsp-venv
+          .lsp-venv/bin/pip install 'ty>=0.0.39,<0.1' boto3 'boto3-stubs[essential]'
+          echo "${{ github.workspace }}/.lsp-venv/bin" >> $GITHUB_PATH
 
       - name: Run LSP integration tests
         run: cargo test -p iam-policy-autopilot-policy-generation --features integ-test -- --ignored
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -203,11 +203,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Install LSP test dependencies
-        run: |
-          python3 -m venv .venv
-          source .venv/bin/activate
-          pip install 'ty>=0.0.39,<0.1' 'boto3-stubs[essential]'
-          echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
+        run: pip install --break-system-packages 'ty>=0.0.39,<0.1' boto3 'boto3-stubs[essential]'
 
       - name: Run LSP integration tests
         run: cargo test -p iam-policy-autopilot-policy-generation --features integ-test -- --ignored
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -209,7 +209,7 @@ jobs:
           echo "${{ github.workspace }}/.lsp-venv/bin" >> $GITHUB_PATH
 
       - name: Run LSP integration tests
-        run: cargo test -p iam-policy-autopilot-policy-generation --features integ-test -- --ignored
+        run: cargo test -p iam-policy-autopilot-policy-generation --features integ-test --test lsp_integration_tests -- --ignored
 
   unused-deps:
     name: Unused Dependencies (cargo-machete)
--- a/iam-policy-autopilot-policy-generation/Cargo.toml
+++ b/iam-policy-autopilot-policy-generation/Cargo.toml
@@ -24,24 +24,22 @@ reqwest.workspace = true
 rust-embed.workspace = true
 schemars.workspace = true
 serde_json.workspace = true
-tokio.workspace = true
 async-trait.workspace = true
 strsim.workspace = true
 derive-new.workspace = true
+lsp-types = "0.95"
+which = "6.0"
+tokio = { workspace = true, features = ["process", "io-util"] }
+tokio-util.workspace = true
 
 
 # Build dependencies
 [build-dependencies]
 serde.workspace = true
-
-# Async runtime
 tokio.workspace = true
 async-trait.workspace = true
 tokio-util.workspace = true
-
-# JSON processing
 serde_json.workspace = true
-
 aws-lc-rs.workspace = true
 git2.workspace = true
 relative-path.workspace = true
@@ -61,6 +59,8 @@ proptest.workspace = true
 test-log = "0.2"
 env_logger.workspace = true
 wiremock = "0.6"
+mockall = "0.12"
+serial_test = "3.0"
 iam-policy-autopilot-policy-generation = { path = ".", features = ["integ-test"]}
 
 # Documentation configuration
--- a/iam-policy-autopilot-policy-generation/Cargo.toml
+++ b/iam-policy-autopilot-policy-generation/Cargo.toml
@@ -28,10 +28,15 @@ reqwest.workspace = true
 rust-embed.workspace = true
 schemars.workspace = true
 serde_json.workspace = true
-tokio.workspace = true
 async-trait.workspace = true
 strsim.workspace = true
 derive-new.workspace = true
+async-lsp = { version = "0.2", features = ["tokio"] }
+lsp-types = "0.95"
+tower = "0.5"
+which = "6.0"
+tokio = { workspace = true, features = ["process", "io-util"] }
+tokio-util.workspace = true
 hcl-rs.workspace = true
 walkdir.workspace = true
 
@@ -39,15 +44,10 @@ walkdir.workspace = true
 [build-dependencies]
 serde.workspace = true
 regex.workspace = true
-
-# Async runtime
 tokio.workspace = true
 async-trait.workspace = true
 tokio-util.workspace = true
-
-# JSON processing
 serde_json.workspace = true
-
 aws-lc-rs.workspace = true
 git2.workspace = true
 relative-path.workspace = true
@@ -70,6 +70,8 @@ paste.workspace = true
 test-log = "0.2"
 env_logger.workspace = true
 wiremock = "0.6"
+serial_test = "3.0"
+temp-env = { version = "0.3", features = ["async_closure"] }
 assert_cmd.workspace = true
 # TLS integration test: verifies the HTTP client respects the OS certificate store
 # (corporate CAs trusted via SSL_CERT_FILE) when connecting to the service reference endpoint
--- a/iam-policy-autopilot-policy-generation/src/lsp/mod.rs
+++ b/iam-policy-autopilot-policy-generation/src/lsp/mod.rs
@@ -1,0 +1,819 @@
+//! LSP (Language Server Protocol) client for Python type information.
+//!
+//! This module provides a minimal, standalone LSP client for communicating with
+//! the [ty](https://github.com/astral-sh/ty) Python type checker. The client uses
+//! JSON-RPC 2.0 over stdin/stdout to query type information via hover requests.
+//!
+//! # Overview
+//!
+//! The LSP client automatically detects and uses ty when available on the system,
+//! requiring no configuration. It manages the ty server process lifecycle, handles
+
+#![allow(dead_code)]
+//! document synchronization, and provides async APIs for querying type information.
+//!
+//! ## Key Features
+//!
+//! - **Automatic ty detection**: Checks PATH for ty availability
+//! - **Process management**: Spawns, initializes, and cleanly shuts down ty server
+//! - **Document handling**: Opens Python files for analysis with proper synchronization
+//! - **Type queries**: Retrieves type information at specific code positions via hover
+//! - **Async-first**: Built on tokio for efficient async I/O operations
+//! - **Error handling**: Comprehensive error types with timeout protection
+//!
+//! # Usage
+//!
+//! ## Basic Example
+//!
+//! ```no_run
+//! use iam_policy_autopilot_policy_generation::lsp::TyLspClient;
+//!
+//! # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+//! // Create and initialize a ty LSP client
+//! let mut client = TyLspClient::new("/path/to/workspace").await?;
+//!
+//! // Open a Python file for analysis
+//! let content = "import boto3\ns3 = boto3.client('s3')";
+//! client.open_document("/path/to/file.py", content).await?;
+//!
+//! // Query type information at a specific position (line 1, character 5)
+//! if let Some(type_info) = client.hover("file:///path/to/file.py", 1, 5).await? {
+//!     println!("Type: {}", type_info);
+//! }
+//!
+//! // Shutdown gracefully
+//! client.shutdown().await?;
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! ## Multiple Documents
+//!
+//! ```no_run
+//! use iam_policy_autopilot_policy_generation::lsp::TyLspClient;
+//!
+//! # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+//! let mut client = TyLspClient::new("/path/to/workspace").await?;
+//!
+//! // Open multiple Python files in the same workspace
+//! client.open_document("/path/to/workspace/app.py", "import boto3").await?;
+//! client.open_document("/path/to/workspace/utils.py", "def helper(): pass").await?;
+//!
+//! // Query type information from any opened document
+//! let type1 = client.hover("file:///path/to/workspace/app.py", 0, 7).await?;
+//! let type2 = client.hover("file:///path/to/workspace/utils.py", 0, 4).await?;
+//!
+//! client.shutdown().await?;
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! ## Error Handling
+//!
+//! ```no_run
+//! use iam_policy_autopilot_policy_generation::lsp::{TyLspClient, LspError};
+//!
+//! # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+//! match TyLspClient::new("/path/to/workspace").await {
+//!     Ok(mut client) => {
+//!         // Use the client
+//!         client.shutdown().await?;
+//!     }
+//!     Err(LspError::TyNotFound) => {
+//!         eprintln!("ty is not installed or not in PATH");
+//!     }
+//!     Err(LspError::Timeout(duration)) => {
+//!         eprintln!("ty server initialization timed out after {:?}", duration);
+//!     }
+//!     Err(e) => {
+//!         eprintln!("Failed to start ty server: {}", e);
+//!     }
+//! }
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! # Public API
+//!
+//! ## Types
+//!
+//! - [`TyLspClient`]: Main client struct for managing ty server and querying type information
+//! - [`LspError`]: Error type for all LSP operations
+//!
+//! # Error Conditions
+//!
+//! The LSP client can fail in several ways:
+//!
+//! ## Startup Errors
+//!
+//! - **[`LspError::TyNotFound`]**: ty command not found in PATH
+//!   - Occurs when ty is not installed or not accessible
+//!   - Check: `which ty` or `ty --version`
+//!
+//! - **[`LspError::StartupFailed`]**: Failed to spawn or initialize ty server
+//!   - Process spawn failed (permissions, missing dependencies)
+//!   - Failed to get stdin/stdout handles
+//!   - Invalid workspace path (non-UTF-8 characters)
+//!
+//! - **[`LspError::InitializeFailed`]**: Initialize request failed
+//!   - ty server rejected initialization
+//!   - ty server returned error response
+//!   - Communication failure during initialization
+//!
+//! ## Communication Errors
+//!
+//! - **[`LspError::Timeout`]**: Operation exceeded timeout duration
+//!   - Initialize: 10 seconds
+//!   - Hover: 5 seconds
+//!   - Shutdown: 2 seconds
+//!   - May indicate ty server is unresponsive or overloaded
+//!
+//! - **[`LspError::SendFailed`]**: Failed to send message to ty server
+//!   - Broken pipe (ty server crashed)
+//!   - I/O error writing to stdin
+//!
+//! - **[`LspError::ParseFailed`]**: Failed to parse LSP response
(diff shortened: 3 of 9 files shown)


iam-policy-autopilot-policy-generation/src/lsp/mod.rs
mschlaipfer, line 1: **[Architecture]** Consider using [`async-lsp`](https://crates.io/crates/async-lsp) (v0.2.4) instead of hand-rolling LSP framing and JSON-RPC dispatch. It handles both client and server mode, process lifecycle (`client-monitor` feature), message routing, and timeouts — which would eliminate most of `protocol.rs`.

More importantly, to support multiple language servers (Go via gopls is a near-term goal), the design should separate transport from server configuration:

```rust
/// Configuration for a specific language server.
pub trait LspServerConfig {
    fn binary_name(&self) -> &str;       // "ty", "gopls"
    fn args(&self) -> &[&str];           // ["server"], ["serve"]
    fn language_id(&self) -> &str;       // "python", "go"
    fn is_available(&self) -> bool;      // which::which check
}

pub struct TyConfig;
impl LspServerConfig for TyConfig {
    fn binary_name(&self) -> &str { "ty" }
    fn args(&self) -> &[&str] { &["server"] }
    fn language_id(&self) -> &str { "python" }
    fn is_available(&self) -> bool { which::which("ty").is_ok() }
}

pub struct GoplsConfig;
impl LspServerConfig for GoplsConfig {
    fn binary_name(&self) -> &str { "gopls" }
    fn args(&self) -> &[&str] { &["serve"] }
    fn language_id(&self) -> &str { "go" }
    fn is_available(&self) -> bool { which::which("gopls").is_ok() }
}

/// Generic LSP client parameterized by server config.
pub struct LspClient<C: LspServerConfig> { /* async-lsp handles transport */ }
```

Adding Go support then becomes ~20 lines (a new config impl) rather than duplicating the entire client.

iam-policy-autopilot-policy-generation/src/lsp/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+//! This module provides a minimal, standalone LSP client for communicating with
+//! the [ty](https://github.com/astral-sh/ty) Python type checker. The client uses
+//! JSON-RPC 2.0 over stdin/stdout to query type information via hover requests.
+//!
+//! # Overview
+//!
+//! The LSP client automatically detects and uses ty when available on the system,
+//! requiring no configuration. It manages the ty server process lifecycle, handles
+
+#![allow(dead_code)]
```
mschlaipfer: **[Code Quality]** Two issues here:

1. `#![allow(dead_code)]` is placed *inside* the `//!` doc block, splitting the module documentation in two. Everything after this line is orphaned from the doc block above.

2. More broadly, both `allow(dead_code)` annotations (here and on the struct at line 232) should be removed. If the module isn't integrated yet, the compiler warning is a useful reminder to wire it in before merge. Suppressing it just hides the fact that nothing calls this code.

Fix: remove both annotations. If you want the module to compile without warnings while WIP, gate it behind a feature flag instead.

iam-policy-autopilot-policy-generation/src/lsp/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+                )))
+            })?;
+
+        // Track the opened document
+        self.opened_documents.insert(file_uri.clone());
+
+        log::debug!("Waiting 1 second for ty to analyze document");
+
+        // Wait 1 second for ty to analyze the document
+        sleep(Duration::from_secs(1)).await;
```
mschlaipfer: **[Design / Testability]** This hardcoded 1s sleep is the dominant cost in integration tests (6 tests × 1s = 6+ seconds of pure waiting). Consider:

1. **Best:** Wait for `textDocument/publishDiagnostics` notification from ty for this URI — that's the signal analysis is complete. You'd need a `pending_messages` buffer for messages consumed while waiting.

2. **Good:** Make the delay configurable (e.g., a `TyLspClientOptions` / builder pattern), defaulting to 1s in production but 0 or short in tests.

3. **Hybrid:** Wait for diagnostics with a short fallback timeout (e.g., 500ms) for servers that skip the notification on clean files.

iam-policy-autopilot-policy-generation/src/lsp/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+    /// 1. Sends a shutdown request with 2s timeout
+    /// 2. Sends an exit notification
+    /// 3. Waits for the process to exit
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - The shutdown request times out
+    /// - Communication with the server fails
+    pub async fn shutdown(mut self) -> Result<(), LspError> {
```
mschlaipfer, line 328: **[Nit]** Duplicate doc comment — there are two `///` doc blocks stacked here (lines 596–603 and 604–615). Keep only the detailed one:

```rust
/// Shutdown the LSP server gracefully.
///
/// This method:
/// 1. Sends a shutdown request with 2s timeout
/// 2. Sends an exit notification
/// 3. Waits for the process to exit
///
/// # Errors
///
/// Returns an error if:
/// - The shutdown request times out
/// - Communication with the server fails
pub async fn shutdown(mut self) -> Result<(), LspError> {
```

iam-policy-autopilot-policy-generation/src/lsp/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+            } else {
+                Some(values.join("\n"))
+            };
+
+            assert_eq!(type_info, Some("Type: str\nAdditional info".to_string()));
+        } else {
+            panic!("Expected array contents");
+        }
+    }
+}
```
mschlaipfer: **[Test Design]** These unit tests re-implement the hover response parsing logic from `hover()` rather than calling the actual code. If the real parsing changes, these tests won't detect regressions.

Suggestion: extract the parsing into a standalone helper:
```rust
fn extract_type_from_hover(response: &serde_json::Value) -> Option<String> { /* ... */ }
```

Then parameterize with rstest (already in dev-deps):
```rust
#[rstest]
#[case(json!({"result": {"contents": "str"}}), Some("str".into()))]
#[case(json!({"result": null}), None)]
#[case(json!({"result": {"contents": null}}), None)]
#[case(json!({"result": {"contents": {"kind": "markdown", "value": "int"}}}), Some("int".into()))]
#[case(json!({"result": {"contents": ["Type: str", {"value": "extra"}]}}), Some("Type: str\nextra".into()))]
fn test_extract_type(#[case] response: Value, #[case] expected: Option<String>) {
    assert_eq!(extract_type_from_hover(&response), expected);
}
```

This tests the real code path and makes each variant a single readable line.

iam-policy-autopilot-policy-generation/src/lsp/protocol.rs
```diff
@@ -0,0 +1,680 @@
...
+///
+/// This function uses `workspace_folders` (the modern LSP approach) for specifying
+/// the workspace. The deprecated `root_uri` field is not set and will be serialized
+/// as `null`, which is correct per the LSP specification when `workspace_folders` is provided.
+#[allow(deprecated)]
+pub fn create_initialize_request(id: u64, workspace_root: &str) -> serde_json::Value {
+    use lsp_types::{ClientCapabilities, InitializeParams, Url, WorkspaceFolder};
+    use serde_json::json;
+
+    let workspace_uri = Url::from_file_path(workspace_root).expect("Invalid workspace path");
```
mschlaipfer: **[Robustness]** `expect()` here (and `Url::parse(...).expect(...)` at line 204) means any invalid input causes a panic rather than a recoverable error. While callers do pre-validate in `TyLspClient::new` and `open_document`, library-internal functions that panic on bad input are surprising.

Consider returning `Result<serde_json::Value, LspError>` from these helpers, or at minimum document the precondition with a `# Panics` doc section so callers know the contract.

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
```diff
@@ -0,0 +1,370 @@
...
+macro_rules! skip_if_no_lsp {
+    () => {
+        if !is_lsp_ready() {
+            eprintln!("Skipping test: ty or boto3-stubs not available");
+            eprintln!("  Install ty: pip install ty");
+            eprintln!("  Install boto3-stubs: pip install 'boto3-stubs[essential|full|all]'");
+            return;
+        }
+    };
+}
```
mschlaipfer: **[Test Reliability]** This macro silently returns early and prints to stderr. In CI, all 6 integration tests will report "passed" while testing nothing — a reviewer can't distinguish "tests ran" from "tests were skipped."

Consider either:
1. Use `#[ignore]` and explicitly run `--ignored` in CI environments where ty is installed
2. Fail loudly in CI:
```rust
macro_rules! skip_if_no_lsp {
    () => {
        if !is_lsp_ready() {
            if std::env::var("CI").is_ok() {
                panic!("LSP integration tests require ty + boto3-stubs in CI");
            }
            eprintln!("Skipping: ty or boto3-stubs not available");
            return;
        }
    };
}
```

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
```diff
@@ -0,0 +1,370 @@
...
+    let method_sig = hover_method.unwrap();
+    assert!(
+        method_sig.contains("list_buckets") && method_sig.contains("ListBucketsOutput"),
+        "Expected list_buckets method signature, got: {}",
+        method_sig
+    );
+
+    // Shutdown gracefully
+    let _: () = client.shutdown().await.expect("Failed to shutdown");
+}
```
mschlaipfer, line 104: **[Test Design]** Every integration test repeats: create workspace → create file → create client → open document → test → shutdown. This is exactly what rstest fixtures solve (and `rstest` is already in dev-deps at line 64 of Cargo.toml!).

With rstest + `#[case]`, the three hover assertions in this function become a single parameterized test:

```rust
#[rstest]
#[case(2, 0, "S3Client")]
#[case(3, 0, "ListBuckets")]
#[case(3, 21, "list_buckets")]
#[tokio::test]
#[serial]
async fn test_hover_returns_expected_type(
    #[case] line: u32,
    #[case] col: u32,
    #[case] expected: &str,
) {
    skip_if_no_lsp!();
    let workspace = TestWorkspace::new().unwrap();
    let file_path = workspace.create_file("test.py", fixtures::SIMPLE_BOTO3).await.unwrap();
    let mut client = TyLspClient::new(&workspace.root_path).await.unwrap();
    client.open_document(&file_path, fixtures::SIMPLE_BOTO3).await.unwrap();

    let hover = client.hover(&workspace.file_uri("test.py"), line, col).await.unwrap();
    assert!(hover.unwrap().contains(expected));
    client.shutdown().await.unwrap();
}
```

Each test case is a single readable line, and adding new assertions is trivial.

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
```diff
@@ -0,0 +1,370 @@
...
+    let file_uri = workspace.file_uri("test.py");
+
+    // Test 1: s3_client variable should return S3Client type
+    eprintln!("\nTest 1: s3_client variable at line 2, char 0");
+    let hover_s3_client: Option<String> = client
+        .hover(&file_uri, 2, 0)
+        .await
+        .expect("Failed to query hover");
+    eprintln!("Result: {:?}", hover_s3_client);
+
```
mschlaipfer: **[Test Readability]** These position numbers (`2, 0`, `3, 0`, `3, 21`) are magic values tightly coupled to the exact text of `fixtures::SIMPLE_BOTO3`. If someone reformats the fixture, these tests break with no obvious reason.

Consider either named constants alongside the fixture:
```rust
// In test_utils/fixtures:
pub const SIMPLE_BOTO3_S3_CLIENT_POS: (u32, u32) = (2, 0);
pub const SIMPLE_BOTO3_RESPONSE_POS: (u32, u32) = (3, 0);
pub const SIMPLE_BOTO3_LIST_BUCKETS_POS: (u32, u32) = (3, 21);
```

Or a helper that computes positions dynamically:
```rust
fn find_position(source: &str, needle: &str) -> (u32, u32) {
    for (line_idx, line) in source.lines().enumerate() {
        if let Some(col) = line.find(needle) {
            return (line_idx as u32, col as u32);
        }
    }
    panic!("'{needle}' not found in fixture");
}
```

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
```diff
@@ -0,0 +1,370 @@
...
+async fn test_ty_not_found_error() {
+    // Temporarily modify PATH to exclude ty
+    let original_path = std::env::var("PATH").unwrap_or_default();
+    std::env::set_var("PATH", "");
+
+    let workspace = TestWorkspace::new().expect("Failed to create workspace");
+    let result: Result<TyLspClient, LspError> = TyLspClient::new(&workspace.root_path).await;
+
+    // Restore original PATH
+    std::env::set_var("PATH", &original_path);
```
mschlaipfer: **[Correctness]** `std::env::set_var` is [unsound in multi-threaded programs](https://doc.rust-lang.org/std/env/fn.set_var.html) and deprecated-as-unsafe since Rust 1.66. Even with `#[serial]`, tokio's worker threads are still running — this is a data race.

Options:
1. Use the [`temp_env`](https://crates.io/crates/temp_env) crate for thread-safe env modification
2. Mark `#[ignore]` and document it must run in isolation (`cargo test -- --ignored --test-threads=1`)
3. Accept that the `TyNotFound` path is already exercised implicitly when `skip_if_no_lsp!()` returns early on machines without ty

iam-policy-autopilot-policy-generation/Cargo.toml
```diff
@@ -67,6 +65,8 @@ rstest.workspace = true
 test-log = "0.2"
 env_logger.workspace = true
 wiremock = "0.6"
+mockall = "0.12"
```
mschlaipfer: **[Nit]** `mockall` is declared here but never used — no `#[automock]` or `mock!` anywhere in the PR.

We should add cargo-machete to our build, opened https://github.com/awslabs/iam-policy-autopilot/issues/206

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
```diff
@@ -0,0 +1,370 @@
...
+
+    eprintln!("Test file content:\n{}", fixtures::SIMPLE_BOTO3);
+
+    // Create and initialize LSP client
+    let mut client: TyLspClient = TyLspClient::new(&workspace.root_path)
+        .await
+        .expect("Failed to create LSP client");
+
+    // Open the document
+    let _: () = client
```
mschlaipfer: **[Nit]** Redundant type annotations throughout the tests:
```rust
let mut client: TyLspClient = TyLspClient::new(...).await...;
let _: () = client.open_document(...).await...;
let hover_s3_client: Option<String> = client.hover(...).await...;
```

The types are obvious from the RHS. `let _: ()` is particularly unusual — just discard the result or use `client.open_document(...).await?;` directly.

mschlaipfer (commented):
Can we use [`async-lsp`](https://crates.io/crates/async-lsp) to avoid hand-rolling the LSP client implementation? Using it would eliminate most of `protocol.rs` and make adding new language servers (e.g., gopls for Go) simpler via a trait-based config. See inline comment for a sketch.

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
mschlaipfer: Follow-up: I checked `.github/workflows/pr-checks.yml` — CI runs `cargo test --workspace` (line 75) but neither `ty` nor `boto3-stubs` are installed in the workflow. So these tests are silently skipped today in CI.

This PR should either:
1. Add a CI step to install the dependencies so the tests actually run:
```yaml
- name: Install LSP test dependencies
  run: |
    pip install ty 'boto3-stubs[essential]'
```
2. Or use `#[ignore]` and add a separate workflow/job that explicitly runs the ignored test with the tools installed.

iam-policy-autopilot-policy-generation/src/lsp/mod.rs
adpaco: Done. Rewrote the entire module using `async-lsp`. The client is generic via `LspServerConfig` trait with `TyConfig` as its first implementation, following the suggested design. Thank you!

iam-policy-autopilot-policy-generation/src/lsp/mod.rs
adpaco: Done. Both annotations have been removed and the module is properly wired in via `pub mod lsp` in the lib.rs

iam-policy-autopilot-policy-generation/src/lsp/mod.rs
adpaco: I implemented "Best" with a fallback behavior using "Good" (it's not so much a delay but a timeout). `open_document` now waits for `publishDiagnostics` for the specific URI. Note that we added URI tracking to avoid races when multiple documents are opened.

iam-policy-autopilot-policy-generation/src/lsp/mod.rs
adpaco: Removed the errors section.

iam-policy-autopilot-policy-generation/src/lsp/mod.rs
adpaco: Thanks, these tests are parametrized now using rstest.

iam-policy-autopilot-policy-generation/src/lsp/protocol.rs
adpaco: Removed (this implementation doesn't exist anymore).

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
adpaco: Ok, these tests have also been parametrized using `rstest`.

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
adpaco: Done. We are now using a helper `find_position` to compute positions dynamically from the fixture text.

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
adpaco: Done, used `temp_env` for this.

iam-policy-autopilot-policy-generation/Cargo.toml
adpaco: Good catch, removed.

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
adpaco: Done.

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
adpaco: Since the LSP is an optional dependency in our case, it made more sense to me to go with option 2. We have `#[ignore]` and a separate job that installs the LSP dependencies and runs the ignored tests. Extending the workflow to test additional LSPs should boil down to adding steps to install the LSPs.

adpaco: Looks like the workflow is failing so I'll mark it as unresolved for now.

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
```diff
@@ -0,0 +1,253 @@
...
+    LspError, TyLspClient,
+};
+use rstest::rstest;
+use serial_test::serial;
+use std::path::PathBuf;
+use tempfile::TempDir;
+use tokio::fs;
+
+struct TestWorkspace {
+    _temp_dir: TempDir,
```
mschlaipfer, line 20: remove unused variable

.github/workflows/pr-checks.yml
```diff
@@ -186,6 +186,31 @@ jobs:
...
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Install LSP test dependencies
+        run: |
+          python3 -m venv .lsp-venv
+          .lsp-venv/bin/pip install 'ty>=0.0.39,<0.1' boto3 'boto3-stubs[essential]'
+          echo "${{ github.workspace }}/.lsp-venv/bin" >> $GITHUB_PATH
+
+      - name: Run LSP integration tests
+        run: cargo test -p iam-policy-autopilot-policy-generation --features integ-test -- --ignored
```
mschlaipfer: Doesn't this need a filter for `lsp_integration_tests`?

iam-policy-autopilot-policy-generation/src/lsp/test_utils.rs
```diff
@@ -0,0 +1,123 @@
+//! Test utilities and fixtures for LSP client testing.
+
+/// Sample Python code fixtures for testing.
+pub mod fixtures {
```
mschlaipfer: Can we move this down, closer to the tests, so it's easier to check against?

iam-policy-autopilot-policy-generation/src/lsp/test_utils.rs
```diff
@@ -0,0 +1,123 @@
...
+        .args(["-c", "import mypy_boto3_s3"])
+        .output()
+        .map(|output| output.status.success())
+        .unwrap_or(false)
+}
+
+/// Check if both ty and boto3-stubs are available.
+#[must_use]
+pub fn is_lsp_ready() -> bool {
+    is_ty_available() && is_boto3_stubs_available()
```
mschlaipfer: All these are Python-specific, can we refactor/rename so we can add support/tests for other languages?

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
```diff
@@ -0,0 +1,253 @@
...
+
+impl TestWorkspace {
+    fn new() -> std::io::Result<Self> {
+        let temp_dir = TempDir::new()?;
+        let root_path = temp_dir.path().to_path_buf();
+
+        // ty resolves third-party imports by looking for a .venv in the project root.
+        // Since test workspaces are temp directories, we write a pyproject.toml pointing
+        // ty to the active Python environment so it can find boto3-stubs.
+        let python_prefix = std::process::Command::new("python3")
```
mschlaipfer, line 32: Same here, could we make it so we can add other languages, or rename to Python-specific files/types?

mschlaipfer (commented):
Thanks for the refactoring!

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
adpaco: It's not unused, but we never read from it. It's there to keep the initialized `TempDir` from being dropped.

.github/workflows/pr-checks.yml
adpaco: Correct, it's been added now.

iam-policy-autopilot-policy-generation/src/lsp/test_utils.rs
adpaco: Done.

iam-policy-autopilot-policy-generation/src/lsp/test_utils.rs
adpaco: Put these below a `python` mod for now. I expect this to be refactored into common helpers when we start adding other LSP clients.

iam-policy-autopilot-policy-generation/tests/lsp_integration_tests.rs
adpaco: See comment above.

mschlaipfer (approved):

janstey-amzn (approved):

Issue #123, opened 2026-01-27
`aws_python_case_conversion` does not correctly translate some operation names


mschlaipfer:
Examples include operations with `Ipv6` and `WhatsApp` in their name.

Instead of special-casing a `snake_case` conversion we should build a map of cases where the standard `snake_case` conversion does not work and override. We can build this map from the service reference https://servicereference.us-east-1.amazonaws.com.

This causes us to miss operations during extraction.

PR #163, opened 2026-03-20
fix: aws_python_case_conversion

mschlaipfer:
*Issue #, if available:* #123

*Description of changes:*

- Adopt the same snake_case conversion logic that [`botocore`](https://github.com/boto/botocore) uses internally for operation and waiter names
- Run this name-conversion logic at **build time** and embed the resulting map into the binary (we filter out all lookups that don't need the custom logic during build to keep the binary small)
- Use the embedded map for both **forward and reverse lookups** at runtime
- Adds Python **linter/formatter check** ([`ruff`](https://docs.astral.sh/ruff/)) to the GitHub Actions workflow
- Document the new **Python dependency** and **style check** requirements
- Clean up existing **tests**

### DIFF
--- a/iam-policy-autopilot-policy-generation/scripts/generate_python_name_map.py
+++ b/iam-policy-autopilot-policy-generation/scripts/generate_python_name_map.py
@@ -40,7 +40,7 @@ def main():
     sys.path.insert(0, botocore_root)
 
     try:
-        from botocore import _xform_cache, xform_name
+        from botocore import xform_name
     except ImportError as e:
         print(
             f"Error: could not import botocore from {botocore_root}: {e}",
--- a/iam-policy-autopilot-policy-generation/scripts/generate_python_name_map.py
+++ b/iam-policy-autopilot-policy-generation/scripts/generate_python_name_map.py
@@ -3,9 +3,7 @@
 Generate a lookup table of AWS operation names → Python snake_case method names.
 
 This script uses botocore's own xform_name() function to produce the authoritative
-mapping for all operation names found in the botocore service data directory, plus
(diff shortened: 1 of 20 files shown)


pyproject.toml
mschlaipfer, line 1: Looks like I just overwrote the existing `pyproject.toml`

iam-policy-autopilot-policy-generation/scripts/generate_python_name_map.py
```diff
@@ -0,0 +1,115 @@
...
+
+    botocore_root = sys.argv[1]
+    botocore_data_dir = sys.argv[2]
+    output_json = sys.argv[3]
+
+    # Add botocore root to sys.path so we can import botocore
+    sys.path.insert(0, botocore_root)
+
+    try:
+        from botocore import _xform_cache, xform_name
```
nwinans-aws: does `xform_name` not already handle `_xform_cache`  interfacing? i think since we already walk the tree for all service-2.json and waiter-2.json files, this shouldn't be necessary.

nwinans-aws (commented):
overall lgtm, i don't think we need the _xform_cache logic though

iam-policy-autopilot-policy-generation/scripts/generate_python_name_map.py
mschlaipfer: nice catch, thanks!

nwinans-aws (dismissed):

nwinans-aws (approved):

weibenz1 (approved):

PR #165, opened 2026-03-23
feat: expand configuration options for service reference cache

### LINKED ISSUE #138: Add Service Reference cache control flags
I'd like to suggest a way for users to control the behaviour of the service reference cache.

Currently, the service reference cache is stored in the OS temp directory, under a `IAMPolicyAutopilot` directory, and expires after 6 hours.

I would suggest the addition of command line flags which could be used to specify an alternative directory, and override the cache expiry.

One use case which this would benefit is when running the command in a docker image. Allowing the user to control the cache directory will enable the user to point it to a volume to ensure the cache can be reused across containers.

It would also benefit those running IAM policy autopilot without internet access or with restrictive firewalls, who could disable the cache expiry.

As an alternative to command line flags, environment variables could be used. Perhaps these would make more sense, particularly in the context of managed environments or docker images.

Additionally, functionality could be added to fallback to an expired cache, with a warning, if internet access is not available.

And yet another option would be just to bundle the service reference into the binary, as is currently done for botocore data. This seems less ideal, but it would remove a lot of complexity, avoiding the need to make web requests and maintain a cache at runtime.

Let me know your thoughts - it depends a lot on the direction you have planned for the project. I'd be happy to implement this feature and make a pull request in whichever way you prefer.

C85297:
Closes #138 

Expands the configuration options for service reference caching. Adds a flag to configure the length of time the cache should be considered valid for, a flag to change the service reference URL, and a flag to configure the location on disk where the cache is stored.

To avoid pulling in more dependencies, for now I made the cache expiry duration string only accept hours, in the format `{n}h`. This enables the potential of adding more complicated duration parsing in the future if desired, without being a breaking change. For now I believe hours should be fine.


iam-policy-autopilot-cli/src/main.rs
```diff
@@ -88,15 +95,35 @@ struct GeneratePolicyCliConfig {
     upload_policies: Option<String>,
     /// Enable minimal policy size by allowing cross-service merging
     minimal_policy_size: bool,
+    /// The URL from which to fetch service reference data
+    service_reference_url: Option<String>,
```
weibenz1, line 99: curious why do we need to externalize this URL?

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -41,6 +43,11 @@ use crate::commands::print_version_info;
 /// Default port for mcp server for Http Transport
 static MCP_HTTP_DEFAULT_PORT: u16 = 8001;
 
+/// Regex for parsing cache expiry duration. For now we only support hours.
+#[allow(clippy::unwrap_used)]
+static CACHE_EXPIRY_DURATION_REGEX: LazyLock<Regex> =
```
weibenz1, line 48: would minutes make more sense?

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -88,15 +95,35 @@ struct GeneratePolicyCliConfig {
     upload_policies: Option<String>,
     /// Enable minimal policy size by allowing cross-service merging
     minimal_policy_size: bool,
+    /// The URL from which to fetch service reference data
+    service_reference_url: Option<String>,
     /// Disable file system caching for service references
     disable_cache: bool,
+    /// File system location for caching service references
+    cache_location: Option<PathBuf>,
+    /// Cache expiry duration string for service references
```
weibenz1, line 104: might want to refine it to be like `in minutes`?

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -304,15 +331,42 @@ for better organization."
...
+        /// Change file system location for caching service references
+        #[arg(
+            long = "cache-location",
+            long_help = "A custom path for file system caching of service reference data. \
+By default, service reference data is cached in the system temp directory. \
+Use this flag to force a different directory to be used."
+        )]
+        cache_location: Option<PathBuf>,
+
+        /// Disable file system caching for service references
```
weibenz1, line 392: seems to be wrong description?

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -452,6 +506,21 @@ async fn handle_generate_policy(config: &GeneratePolicyCliConfig) -> Result<()>
             service_names: names.clone(),
         });
 
+    let cache_expiry_seconds = if let Some(cache_expiry) = &config.cache_expiry {
```
weibenz1, line 603: should this logic be moved inside service reference since that's the only place cache is managed? It will make the parameter validation easier as well - e.g. validating --disable-cache & other cache related params

weibenz1 (commented):
thanks for the contribution!

PR #174, opened 2026-04-07
feat: anonymous telemetry 

weibenz1:
*Issue #, if available:*

*Description of changes:*

This PR introduces client side telemetry framework for helping collecting data to improve IPA's security and accuracy.

No PII or customer content is collected. More details can be found in TELEMETRY.md.


iam-policy-autopilot-cli/Cargo.toml
```diff
@@ -6,6 +6,7 @@ description = "Standalone CLI for AWS IAM Policy Autopilot: analyzes AccessDenie
 
 [dependencies]
 iam-policy-autopilot-access-denied = { path = "../iam-policy-autopilot-access-denied" }
+iam-policy-autopilot-common = { path = "../iam-policy-autopilot-common" }
```
mschlaipfer, line 9: We already have the https://github.com/awslabs/iam-policy-autopilot/blob/main/iam-policy-autopilot-tools crate which we should use for this. Feel free to rename `...-tools` is not great. `-common` or `-utils` would be better.

iam-policy-autopilot-common/src/telemetry/config.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+/// Get the path to the `.iam-policy-autopilot` directory.
+fn dirs_path() -> Option<PathBuf> {
+    home_dir().map(|home| home.join(CONFIG_DIR_NAME))
+}
+
+/// Get the user's home directory.
+///
+/// Uses `$HOME` on Unix/macOS and `%USERPROFILE%` on Windows.
```
mschlaipfer, line 92: is this tested somewhere?

iam-policy-autopilot-common/src/telemetry/config.rs
```diff
@@ -0,0 +1,295 @@
...
+/// a new UUID is generated and persisted automatically.
+#[must_use]
+pub fn installation_id() -> String {
+    cached_config().installation_id.clone()
+}
+
+/// Update the telemetry choice in the persistent config file.
+///
+/// Writes through to disk immediately. Note: the in-process cache retains
+/// the original value (acceptable since the CLI exits shortly after this call).
```
mschlaipfer, line 181: Could we update the in-process config, so we don't risk breaking this invariant down the road?

iam-policy-autopilot-common/src/telemetry/config.rs
```diff
@@ -0,0 +1,295 @@
...
+        let config = TelemetryConfig {
+            installation_id: "550e8400-e29b-41d4-a716-446655440000".to_string(),
+            telemetry_choice: TelemetryChoice::Enabled,
+        };
+        let json = serde_json::to_string_pretty(&config).unwrap();
+        for needle in [
+            "\"installationId\"",
+            "\"telemetryChoice\"",
+            "\"enabled\"",
+            "550e8400",
```
mschlaipfer: ?

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
```diff
@@ -241,8 +241,6 @@ pub async fn generate_policies(config: &GeneratePolicyConfig) -> Result<Generate
     .await
     .context("Failed to process source files")?;
 
-    // Relies on the invariant that all source files must be of the same language, which we
-    // enforce in process_source_files
```
mschlaipfer, line 245: why is this deleted? This is still true, we take `.first()`.

iam-policy-autopilot-policy-generation/src/extraction/engine.rs
```diff
@@ -143,6 +143,12 @@ impl Engine {
         // Disambiguate and validate method calls against SDK definitions
         extractor.disambiguate(&mut all_extraction_results, &service_index);
 
+        // Record detected language into telemetry span (no-ops if no scope active)
+        iam_policy_autopilot_common::telemetry::span::record_result_str(
+            "detected_language",
+            &language.to_string(),
+        );
```
mschlaipfer: This should be done here instead: https://github.com/awslabs/iam-policy-autopilot/blob/main/iam-policy-autopilot-policy-generation/src/api/common.rs#L34-L48

iam-policy-autopilot-policy-generation/src/policy_generation/engine.rs
```diff
@@ -63,6 +63,12 @@ impl<'a> Engine<'a> {
         for enriched_call in enriched_calls {
             let policy = self.generate_policy_for_call(enriched_call)?;
             policies.push(policy);
+
+            // Record service into telemetry span (no-ops if no scope active)
+            iam_policy_autopilot_common::telemetry::span::record_result_set(
+                "services_used",
+                &enriched_call.service,
+            );
```
mschlaipfer: This should be done here instead: https://github.com/awslabs/iam-policy-autopilot/blob/main/iam-policy-autopilot-policy-generation/src/enrichment/engine.rs#L54

iam-policy-autopilot-proc-macros/src/telemetry.rs
mschlaipfer, line 1: This looks neat, but I'm worried that we are implementing our own telemetry framework. Is there no existing project we could use instead?

TELEMETRY.md
```diff
@@ -0,0 +1,177 @@
+# Telemetry
+
+IAM Policy Autopilot collects **anonymous, non-personally-identifiable** usage telemetry to help the development team understand how the tool is used, prioritize feature development, and measure adoption.
+
+## What Is Collected
+
+Telemetry records **only** which commands and parameters are used, and whether the command succeeded. It **never** collects file paths, file contents, AWS account IDs, AWS regions, credentials, policy content, or any personally identifiable information.
+
+<!-- BEGIN AUTO-GENERATED TELEMETRY TABLE -->
```
mschlaipfer: How does this work? Do we generate this in CI and check equivalence to the table here?

iam-policy-autopilot-common/src/telemetry/span.rs
```diff
@@ -0,0 +1,555 @@
...
+        assert!(result.is_ok());
+    }
+
+    // =========================================================================
+    // TELEMETRY.md validation — every record_result_* key in production code
+    // must be documented in the "Result Data" section of TELEMETRY.md.
+    // =========================================================================
+
+    #[test]
+    fn test_span_result_fields_documented_in_telemetry_md() {
```
mschlaipfer: This should be in its own file.

iam-policy-autopilot-common/src/telemetry/span.rs
```diff
@@ -0,0 +1,555 @@
...
+            .filter(|e| {
+                let path = e.path();
+                path.extension().is_some_and(|ext| ext == "rs")
+                    && !path.to_str().unwrap_or("").contains("/target/")
+                    && !path.to_str().unwrap_or("").contains("/tests/")
+            })
+        {
+            let content = std::fs::read_to_string(entry.path()).unwrap_or_default();
+
+            // Truncate at #[cfg(test)] to exclude inline test modules
```
mschlaipfer: This is brittle and won't work generally, I can write `#[cfg(test)]` at the top of a file.

iam-policy-autopilot-common/tests/telemetry_derive_test.rs
```diff
@@ -0,0 +1,562 @@
...
+            None => std::env::remove_var(TELEMETRY_ENV_VAR),
+        }
+    }
+}
+
+// =============================================================================
+// Test fixtures (types)
+// =============================================================================
+
+#[allow(dead_code)]
```
mschlaipfer: Why is this dead code?

iam-policy-autopilot-common/tests/telemetry_derive_test.rs
```diff
@@ -0,0 +1,562 @@
...
+    // Unit variant
+    #[telemetry(command = "help")]
+    Help,
+
+    // Tuple variant (skip)
+    #[telemetry(skip)]
+    Internal(String),
+}
+
+#[allow(dead_code)]
```
mschlaipfer: Why is this dead code?

mschlaipfer (changes requested):

iam-policy-autopilot-common/src/telemetry/config.rs
weibenz1: could you elaborate? what kind of of update are you referring to?

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
weibenz1: might be a rebase, will revert this change

iam-policy-autopilot-common/src/telemetry/config.rs
mschlaipfer: I thought to defensively program here, but now reading the code it seems unproblematic due to taking ownership of `choice`.

iam-policy-autopilot-common/src/telemetry/config.rs
weibenz1: added a dedicated test in next revision

iam-policy-autopilot-cli/Cargo.toml
weibenz1: cool, will refactor

TELEMETRY.md
weibenz1: yes, there are tests in both cli and mcp crate, and it will fail if there's any mismatch when the tests are executed.

iam-policy-autopilot-proc-macros/src/telemetry.rs
weibenz1: per discussed offline, opentelemetry is for distributed tracing along with things to prevent vendor lockin. It's mostly also used for service with a runtime framework that manages request/response concepts - like interceptors in Java, which will be quite an overkill for us.

iam-policy-autopilot-common/tests/telemetry_derive_test.rs
weibenz1: this is a test file, and the file is not directly accessed - it's to verify that the telemetry(skip) macros function as is.

iam-policy-autopilot-common/tests/telemetry_derive_test.rs
weibenz1: same as above

iam-policy-autopilot-common/src/telemetry/span.rs
weibenz1: addressed in next revision - moved to use syn crate for parsing.

iam-policy-autopilot-common/src/telemetry/span.rs
weibenz1: sure, moved

iam-policy-autopilot-cli/Cargo.toml
weibenz1: actually, it's not a small refactor. tools has a lot dependency on policy-generation, it would create circular dependency if we moved it just as it; and the refactoring would require many updates in the code base; I will create an issue for it.

weibenz1: https://github.com/awslabs/iam-policy-autopilot/issues/176

iam-policy-autopilot-common/tests/telemetry_derive_test.rs
mschlaipfer: Can we annotate just the fields so this is more obvious, and not the entire type, or add a comment to the annotation if we keep it at the type level? This type is used in this very file, so the placement of the annotation confused me.

iam-policy-autopilot-mcp-server/src/tools/generate_policy.rs
```diff
@@ -196,7 +196,10 @@ mod tests {
             tfvars: None,
         };
 
-        let expected_output = include_str!("../testdata/test_generate_application_policy");
+        // Normalize line endings: on Windows, Git may check out with \r\n,
+        // but serde_json::to_string_pretty always uses \n.
+        let expected_output =
+            include_str!("../testdata/test_generate_application_policy").replace('\r', "");
```
mschlaipfer: how would this have worked so far?

mschlaipfer: Should we configure the Github workflow with `git config --system core.autocrlf false` (we already configure `longpaths`)? It seems like the issue we're trying to resolve is just for equality comparison in tests, and this configuration wouldn't make us miss bugs in Windows. If we do that we should add it here: https://github.com/awslabs/iam-policy-autopilot/blob/main/CONTRIBUTING.md#windows

weibenz1: we used to only test in linux before. Now, given we are adding more filepath related functionalities, i've enabled it for all OS & architecture in this PR.

yeah, probably easier to set gitconfig

weibenz1: added to .gitattributes. that way, it's automatically set for all CI jobs, as well as any clone - hence on need to update contributing.md. Lets see if it passes.

iam-policy-autopilot-common/tests/telemetry_derive_test.rs
weibenz1: addressed in next commit

iam-policy-autopilot-common/src/telemetry/client.rs
```diff
@@ -0,0 +1,246 @@
...
+//! All errors are silently ignored to ensure telemetry never impacts tool functionality.
+
+use log::debug;
+use reqwest::Client;
+use sha2::{Digest, Sha256};
+
+use super::event::TelemetryEvent;
+
+/// The CloudFront distribution endpoint for telemetry ingestion.
+const TELEMETRY_ENDPOINT: &str = "https://d3l5r34trbj0zx.cloudfront.net/";
```
adpaco, line 19: We should consider a custom domain to be more transparent. It'd also make it easier to block through DNS.

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -368,6 +387,7 @@ Examples:\n  \
 --explain 'ec2:Describe*'     # Explain EC2 Describe actions\n  \
 --explain 's3:*' 'dynamodb:*' # Explain S3 and DynamoDB actions"
         )]
+        #[telemetry(list)]
         explain: Option<Vec<String>>,
```
adpaco, line 391: The `#[telemetry(list)]` on `service_hints` and `explain` could be an issue because the contents would be recorded verbatim. For `service_hints` we could validate before the event, but `explain` uses patterns so we should probably do something else.

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -636,7 +706,11 @@ async fn main() {
                 Some(text) => text,
             };
 
-            commands::fix_access_denied(&error_text, yes).await
+            Box::pin(telemetry::span::run_with_telemetry(
+                commands::fix_access_denied(&error_text, yes),
+                &mut telemetry_event,
```
adpaco, line 711: Why is this needed?

iam-policy-autopilot-mcp-server/src/mcp.rs
```diff
@@ -77,15 +82,19 @@ impl IamAutoPilotMcpServer {
         params: Parameters<GeneratePoliciesInput>,
     ) -> Result<Json<GeneratePoliciesOutput>, McpError> {
         trace!("generate_application_policies input: {:#?}", params.0);
+        let telemetry_event = params.0.to_telemetry_event();
 
-        let output = generate_application_policies(params.0).await.map_err(|e| {
+        Box::pin(run_with_telemetry_emit(
```
adpaco, line 87: Why is this needed?

iam-policy-autopilot-common/src/telemetry/config.rs
```diff
@@ -0,0 +1,321 @@
...
+    cached_config().installation_id.clone()
+}
+
+/// Update the telemetry choice in the persistent config file.
+///
+/// Writes through to disk immediately. Note: the in-process cache retains
+/// the original value (acceptable since the CLI exits shortly after this call).
+pub fn set_telemetry_choice(choice: TelemetryChoice) {
+    let mut config = cached_config().clone();
+    config.telemetry_choice = choice;
```
adpaco, line 184: Why use a OnceLock and not RwLOck?

iam-policy-autopilot-common/src/telemetry/client.rs
weibenz1: this can be a follow up task if we really want it. Cloudfront already provides good layers of protection for us together with WAF

iam-policy-autopilot-cli/src/main.rs
weibenz1: these are generic service names/actions tho - e.g. `s3, iam:*` - are you worried about users providing something else other than expected?

iam-policy-autopilot-common/src/telemetry/config.rs
weibenz1: it's write once with many reads - especially for mcp workflows. RW will cause overhead for the reads.

iam-policy-autopilot-cli/src/main.rs
weibenz1: This is a pretty common practice to prevent large futures occupying too much stack space; instead, with Box:pin, it's allocated to a fixed heap space without occupying stack. It's actually flagged by our clippy check.

iam-policy-autopilot-policy-generation/tests/terraform_integration.rs
mschlaipfer, line 75: The problem arises here.
```
rstest #[files] glob
    │
    ▼  canonicalize() internally
fixture_json = \\?\D:\a\iam-policy-autopilot\...\state_precedence\fixture.json
    │
    ▼  .parent().unwrap()
fixture_dir = \\?\D:\a\iam-policy-autopilot\...\state_precedence\
    │
    ▼  fixture_dir.join("main.tf")  ← in build_config(), line 61
terraform_files[0] = \\?\D:\a\iam-policy-autopilot\...\state_precedence\main.tf
    │
    ▼  flows into GeneratePolicyConfig.terraform_files
    ▼  → hcl_parser → Location::new(source_path.to_path_buf(), ...)
    ▼  → Location.file_path = \\?\D:\...\main.tf
    │
    ▼  to_gnu_format() → replace('\\', '/')
"///?/D:/a/iam-policy-autopilot/.../main.tf:13.1-22.2"

```

iam-policy-autopilot-policy-generation/src/lib.rs
```diff
@@ -243,10 +243,16 @@ impl Location {
...
     #[must_use]
     pub fn to_gnu_format(&self) -> String {
-        // Use forward slashes universally — PathBuf::display() uses `\` on Windows
-        let path_str = self.file_path.to_string_lossy().replace('\\', "/");
+        let raw = self.file_path.to_string_lossy();
+        // Strip the Windows extended-length path prefix (\\?\) — it's an internal
+        // Windows API detail, not meaningful in GNU-format user-facing output.
+        let cleaned = raw.strip_prefix(r"\\?\").unwrap_or(&raw);
+        // Use forward slashes universally — PathBuf uses `\` on Windows
+        let path_str = cleaned.replace('\\', "/");
```
mschlaipfer: I don't think this is the right way to fix this.

This can only happen in our tests, and not in production code, due to how `rstest` works with the `#[files(...)]` glob. We don't populate `Location::file_path` with a canonicalized `PathBuf`. The comment on lines 246-247 only applies to test code. See my comment in `build_config()`.

iam-policy-autopilot-policy-generation/tests/terraform_integration.rs
mschlaipfer, line 210: One IMHO better way to fix it would be to strip the prefix here, or to encode the path within the fixture itself rather than joining the path to the canonicalized path in `build_config`.

mschlaipfer (changes requested):

iam-policy-autopilot-policy-generation/src/lib.rs
weibenz1: thanks, good catch!

iam-policy-autopilot-policy-generation/tests/terraform_integration.rs
weibenz1: fixed in new commit

iam-policy-autopilot-common/tests/telemetry_doc_sync_test.rs
```diff
@@ -0,0 +1,217 @@
...
+    );
+
+    for key in &recorded_keys {
+        assert!(
+            result_section.contains(&format!("| `{key}` |")),
+            "TELEMETRY.md Result Data section missing documentation for `{key}`. \
+             Every record_result_* key must have a row between the \
+             BEGIN/END RESULT DATA TABLE markers."
+        );
+    }
```
mschlaipfer, line 176: Should we also check the other direction, so that we don't document something that we've removed from code?

mschlaipfer (commented):
Unless I missed it, it would be nice to see a top-level integration test, which runs with telemetry and has an `assert_eq!` with an expected JSON file, which shows what we produce.

iam-policy-autopilot-common/src/telemetry/client.rs
adpaco: It's not about protecting us, but being more transparent and manageable for users. If they inspect their network traffic and see this address they won't be able to recognize it.

adpaco: I'd recommend creating a follow-up task for this.

iam-policy-autopilot-cli/src/main.rs
adpaco: Yes, they could accidentally provide arbitrary strings which would be recorded as is.

weibenz1:
> Unless I missed it, it would be nice to see a top-level integration test, which runs with telemetry and has an `assert_eq!` with an expected JSON file, which shows what we produce.

do you mean to test telemetry? we don't output anything in the output for telemetry tho. Could you elaborate a bit?

mschlaipfer:
Ah right, somehow I didn't think this through. Never mind.

iam-policy-autopilot-common/tests/telemetry_doc_sync_test.rs
weibenz1: good point, updated in new commit.

iam-policy-autopilot-mcp-server/Cargo.toml
```diff
@@ -34,13 +34,16 @@ rmcp = { version = "^0.8" , features = [
...
+iam-policy-autopilot-common = { path = "../iam-policy-autopilot-common" }
 iam-policy-autopilot-policy-generation = { path = "../iam-policy-autopilot-policy-generation" }
 iam-policy-autopilot-access-denied = { path = "../iam-policy-autopilot-access-denied" }
 
 [dev-dependencies]
 iam-policy-autopilot-policy-generation = { path = "../iam-policy-autopilot-policy-generation" , features=["integ-test"] }
+iam-policy-autopilot-common = { path = "../iam-policy-autopilot-common" }
 serial_test.workspace = true
 reqwest.workspace = true
+rstest = "0.24"
```
mschlaipfer:
```suggestion
rstest.workspace = true
```

weibenz1: resolved in new commit

mschlaipfer (approved):

adpaco-aws (approved):

iam-policy-autopilot-cli/src/main.rs
weibenz1: It seems to be out of our boundary to considering we've disclosed that we record value for these, and the expected format for these values are strings like s3:*.

PR #179, opened 2026-04-24
Integrate with service reference modified timestamp for smarter cache control

### LINKED ISSUE #175: Service Reference: integration with modified field
service reference is introducing a lastModified field, which we can use for refreshing cache.

nwinans-aws:
## Integrate with service reference modified timestamp for smarter cache control

Fixes #175

### Problem

The service reference loader uses a 5-minute TTL to expire cached service data. This means:
- Every 5 minutes, all cached services are re-fetched regardless of whether they changed
- Between refreshes, stale data can be served even if the service reference has updated

The service reference recently added a `modified` epoch timestamp to each entry in its index at `servicereference.us-east-1.amazonaws.com`, enabling clients to know exactly when a service's auth model changed.

### Solution

Replace the TTL-based cache with `modified`-timestamp-based invalidation:

1. **Once per invocation**, `refresh_index()` fetches the lightweight index and compares each service's `modified` against the locally cached version. Stale entries are evicted from both in-memory and filesystem caches.
2. **During enrichment**, individual services are lazily loaded only when needed — unchanged services are never re-fetched.
3. **Optional throttle**: Set `IAM_POLICY_AUTOPILOT_REFRESH_INTERVAL_SECONDS` to skip the index refresh if it was already fetched within that window. Unset (default) refreshes every invocation.

### Changes

**`service_reference.rs`**
- Added `ServiceReferenceEntry` with `url` + `modified` (u64 epoch)
- Replaced `OnceCell` with `RwLock<Option<>>` so the index can be re-fetched
- `service_cache` tracks `modified` timestamp instead of `SystemTime`
- Added `refresh_index()` — fetches index, evicts stale entries
- Added `fetch_mapping()` — extracted HTTP fetch logic
- `load()` uses `modified`-based validity; lazily inits mapping for callers outside the enrichment path (e.g. Terraform resource resolution)
- Filesystem cache uses `.modified` sidecar files instead of file mtime
- Removed `DEFAULT_CACHE_DURATION_IN_SECONDS` and all TTL logic

**`engine.rs`**
- Calls `refresh_index()` (best-effort) once before the enrichment loop
- Transient network failures fall back to cached/lazy-loaded data

**`mod.rs`**
- Mock server index responses now include `modified`
- Mock setup functions call `refresh_index()` to populate the mapping

**`resource_matcher.rs`**
- Tests that mount additional service mocks call `refresh_index()` afterward

### Testing

- All 695 existing tests pass (693 original + 2 new)
- 2 new integration tests using wiremock:
  - `test_modified_based_cache_invalidation`: verifies same-modified keeps cache, bumped-modified evicts and triggers re-fetch
  - `test_refresh_throttle_behavior`: verifies env var throttle skip/allow/zero semantics
- 2 existing `#[ignore]` integration tests pass against the real service reference endpoint

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
@@ -68,6 +68,9 @@ impl Engine {
             .await?;
 
         let resource_matcher = ResourceMatcher::new(service_cfg, fas_maps, sdk);
+        if let Err(e) = self.service_reference_loader.refresh_index().await {
+            log::warn!("Failed to refresh service reference index, will use cached data: {e}");
+        }
         let enriched_calls = self
             .enrich_all_methods(extracted_methods, &resource_matcher)
             .await?;
--- a/iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
@@ -68,6 +68,10 @@ impl Engine {
             .await?;
 
         let resource_matcher = ResourceMatcher::new(service_cfg, fas_maps, sdk);
+        // Refresh the service reference index once before the enrichment loop
+        // so stale cache entries are evicted up front. Individual services are
+        // then lazily fetched by load() only when actually needed during
+        // enrichment — unchanged services are never re-downloaded.
         if let Err(e) = self.service_reference_loader.refresh_index().await {
             log::warn!("Failed to refresh service reference index, will use cached data: {e}");
         }
--- a/iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
@@ -72,6 +72,13 @@ impl Engine {
 
         let resource_matcher =
             ResourceMatcher::new(service_cfg, fas_maps, sdk, self.resource_cutoff);
+        // Refresh the service reference index once before the enrichment loop
+        // so stale cache entries are evicted up front. Individual services are
+        // then lazily fetched by load() only when actually needed during
+        // enrichment — unchanged services are never re-downloaded.
+        if let Err(e) = self.service_reference_loader.refresh_index().await {
+            log::warn!("Failed to refresh service reference index, will use cached data: {e}");
+        }
         let enriched_calls = self
             .enrich_all_methods(extracted_methods, &resource_matcher)
             .await?;
--- a/iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
@@ -592,8 +592,8 @@ pub(crate) mod mock_remote_service_reference {
         Mock::given(method("GET"))
             .and(path("/"))
             .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!([
-                {"service": "s3", "url": format!("{}/s3.json", mock_server_url)},
-                {"service": service_name, "url": format!("{}/{}.json", mock_server_url, service_name)}
+                {"service": "s3", "url": format!("{}/s3.json", mock_server_url), "modified": 1000},
+                {"service": service_name, "url": format!("{}/{}.json", mock_server_url, service_name), "modified": 1000}
             ])))
             .with_priority(1)
             .mount(mock_server)
@@ -615,7 +615,7 @@ pub(crate) mod mock_remote_service_reference {
         Mock::given(method("GET"))
             .and(path("/"))
             .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!([
-                {"service": "s3", "url": format!("{}/s3.json", mock_server_url)}
+                {"service": "s3", "url": format!("{}/s3.json", mock_server_url), "modified": 1000}
             ])))
             .mount(&mock_server)
             .await;
@@ -670,6 +670,7 @@ pub(crate) mod mock_remote_service_reference {
         let loader = RemoteServiceReferenceLoader::new(true)
             .unwrap()
             .with_mapping_url(mock_server_url);
+        loader.refresh_index().await.unwrap();
 
         (mock_server, loader)
     }
@@ -685,7 +686,7 @@ pub(crate) mod mock_remote_service_reference {
         Mock::given(method("GET"))
             .and(path("/"))
             .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!([
-                {"service": "s3", "url": format!("{}/s3.json", mock_server_url)}
+                {"service": "s3", "url": format!("{}/s3.json", mock_server_url), "modified": 1000}
             ])))
             .mount(&mock_server)
             .await;
@@ -801,6 +802,7 @@ pub(crate) mod mock_remote_service_reference {
         let loader = RemoteServiceReferenceLoader::new(true)
             .unwrap()
             .with_mapping_url(mock_server_url);
+        loader.refresh_index().await.unwrap();
 
         (mock_server, loader)
     }
--- a/iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
@@ -760,8 +760,8 @@ pub(crate) mod mock_remote_service_reference {
         Mock::given(method("GET"))
             .and(path("/"))
             .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!([
-                {"service": "s3", "url": format!("{}/s3.json", mock_server_url)},
-                {"service": service_name, "url": format!("{}/{}.json", mock_server_url, service_name)}
+                {"service": "s3", "url": format!("{}/s3.json", mock_server_url), "modified": 1000},
+                {"service": service_name, "url": format!("{}/{}.json", mock_server_url, service_name), "modified": 1000}
             ])))
             .with_priority(1)
             .mount(mock_server)
@@ -783,7 +783,7 @@ pub(crate) mod mock_remote_service_reference {
         Mock::given(method("GET"))
             .and(path("/"))
             .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!([
-                {"service": "s3", "url": format!("{}/s3.json", mock_server_url)}
+                {"service": "s3", "url": format!("{}/s3.json", mock_server_url), "modified": 1000}
             ])))
             .mount(&mock_server)
             .await;
@@ -838,6 +838,7 @@ pub(crate) mod mock_remote_service_reference {
         let loader = RemoteServiceReferenceLoader::new(true)
             .unwrap()
             .with_mapping_url(mock_server_url);
+        loader.refresh_index().await.unwrap();
 
         (mock_server, loader)
     }
@@ -851,7 +852,7 @@ pub(crate) mod mock_remote_service_reference {
         Mock::given(method("GET"))
             .and(path("/"))
             .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!([
-                {"service": "s3", "url": format!("{}/s3.json", mock_server_url)}
+                {"service": "s3", "url": format!("{}/s3.json", mock_server_url), "modified": 1000}
             ])))
             .mount(&mock_server)
             .await;
@@ -967,6 +968,7 @@ pub(crate) mod mock_remote_service_reference {
         let loader = RemoteServiceReferenceLoader::new(true)
             .unwrap()
             .with_mapping_url(mock_server_url);
+        loader.refresh_index().await.unwrap();
 
         (mock_server, loader)
     }
--- a/iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
@@ -16,13 +16,15 @@ use std::{
     time::{Duration, SystemTime},
 };
 use tokio::fs;
-use tokio::sync::{OnceCell, RwLock};
+use tokio::sync::RwLock;
 
 type OperationName = String;
 const IAM_POLICY_AUTOPILOT: &str = "IAMPolicyAutopilot";
-// Cache files for 5 minutes.
-// We can allow cache duration override in future.
-const DEFAULT_CACHE_DURATION_IN_SECONDS: u64 = 300;
+
+/// Environment variable to throttle index refreshes. When set to a number of
+/// seconds, `refresh_index()` becomes a no-op if the index was already fetched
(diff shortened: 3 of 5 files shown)


iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -349,65 +370,112 @@ impl RemoteServiceReferenceLoader {
...
+
+    /// Refresh the service index and invalidate any cached entries whose
+    /// `modified` timestamp has increased. Call once per invocation, before
+    /// the enrichment loop.
+    ///
+    /// When `IAM_POLICY_AUTOPILOT_REFRESH_INTERVAL_SECONDS` is set, the
+    /// refresh is skipped if the index was already fetched within that window.
+    pub(crate) async fn refresh_index(&self) -> crate::errors::Result<()> {
+        // Check optional throttle
+        if let Ok(val) = std::env::var(REFRESH_INTERVAL_ENV) {
```
adpaco: Instead of reading `std::env::var` directly, could we have the refresh interval be a configuration field of the `RemoteServiceReferenceLoader` (e.g., an `Option<Duration>` set at construction through the env. variable) and provide methods to set/unset the value? This would allow us to avoid the `set_var`/`revmove_var` in the tests which are not thread-safe.

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -604,37 +693,37 @@ mod tests {
...
         if let Some(entry) = loader.service_cache.write().await.get_mut("s3") {
-            entry.1 = expired_time;
+            entry.1 = 0; // very old timestamp
         }
 
-        // Load again - should fetch fresh data, not use expired cache
+        // refresh_index should evict the stale entry since mock returns modified=0 by default
+        // but our cached value is also 0, so it won't evict. Let's update the mapping to have
+        // a higher modified value by re-fetching (the mock always returns the same).
+        // Instead, verify that load() still works after cache is populated.
```
adpaco: This comment is confusing. We've done some setup to assert that `load()` works as expected?

adpaco-aws (commented):
Thank you, @nwinans-aws ! It's looking good, I only have a few comments. The PR description says it fixes #75 but that's another PR? Or what's the connection to it?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -264,21 +266,32 @@ where
...
 
 fn deserialize_service_reference_mapping(
     value: Value,
-) -> crate::errors::Result<HashMap<String, Url>> {
+) -> crate::errors::Result<HashMap<String, ServiceReferenceEntry>> {
     #[derive(Deserialize)]
     struct ServiceEntry {
         service: String,
         url: String,
+        #[serde(default)]
```
mschlaipfer, line 291: Why do we need `default`? Does this also serialize cached data (since this doesn't derive `Serialize` I assume not)?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -305,8 +324,12 @@ fn deserialize_service_reference_mapping(
 
 pub(crate) struct RemoteServiceReferenceLoader {
     client: Client,
-    service_reference_mapping: OnceCell<ServiceReferenceMapping>,
-    service_cache: RwLock<HashMap<String, (ServiceReference, SystemTime)>>,
+    /// The index mapping; refreshed by `refresh_index()` before each enrichment run.
+    service_reference_mapping: RwLock<Option<ServiceReferenceMapping>>,
+    /// In-memory cache keyed by service name → (data, modified timestamp from index).
+    service_cache: RwLock<HashMap<String, (ServiceReference, u64)>>,
+    /// Tracks when the index was last refreshed, for optional throttling.
```
mschlaipfer: "for optional throttling"?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -349,65 +370,112 @@ impl RemoteServiceReferenceLoader {
...
+    }
+
+    /// Refresh the service index and invalidate any cached entries whose
+    /// `modified` timestamp has increased. Call once per invocation, before
+    /// the enrichment loop.
+    ///
+    /// When `IAM_POLICY_AUTOPILOT_REFRESH_INTERVAL_SECONDS` is set, the
+    /// refresh is skipped if the index was already fetched within that window.
+    pub(crate) async fn refresh_index(&self) -> crate::errors::Result<()> {
+        // Check optional throttle
```
mschlaipfer: Why do we call this "throttle"? We just check if we need to refresh right?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -448,15 +516,14 @@ impl RemoteServiceReferenceLoader {
...
-        if let Ok(metadata) = fs::metadata(path).await {
-            if let Ok(modified) = metadata.modified() {
-                if let Ok(elapsed) = SystemTime::now().duration_since(modified) {
-                    return elapsed < Duration::from_secs(DEFAULT_CACHE_DURATION_IN_SECONDS);
-                }
-            }
-        }
-        false
+    fn get_cache_meta_path(service_name: &str) -> PathBuf {
+        Self::get_cache_dir().join(format!("{service_name}.modified"))
```
mschlaipfer, line 549: Any downsides having one file per service?

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -801,6 +802,7 @@ pub(crate) mod mock_remote_service_reference {
         let loader = RemoteServiceReferenceLoader::new(true)
             .unwrap()
             .with_mapping_url(mock_server_url);
+        loader.refresh_index().await.unwrap();
```
mschlaipfer, line 971: Why do we now need to explicitly call this? Seems like this could be easy to miss?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -349,65 +370,112 @@ impl RemoteServiceReferenceLoader {
...
+            cache.remove(name);
+            if !self.disable_file_system_cache {
+                let _ = fs::remove_file(Self::get_cache_path(name)).await;
+                let _ = fs::remove_file(Self::get_cache_meta_path(name)).await;
+            }
+        }
+        drop(cache);
+
+        *self.service_reference_mapping.write().await = Some(new_mapping);
+        *self.last_refreshed.write().await = Some(SystemTime::now());
```
mschlaipfer, line 505: Does this correspond to `cached_modified`? I find this code hard to follow, because the logic to evict and write new entries looks very different.

iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
```diff
@@ -68,6 +68,9 @@ impl Engine {
             .await?;
 
         let resource_matcher = ResourceMatcher::new(service_cfg, fas_maps, sdk);
+        if let Err(e) = self.service_reference_loader.refresh_index().await {
```
mschlaipfer, line 79: Why do we need this explicit `refresh_index()` now?

Also, and probably unrelated to this PR: Where do we ensure that we only download the service reference for operations found by static analysis/FAS expansion?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -604,37 +693,37 @@ mod tests {
...
         let result = loader.load("s3").await;
         assert!(result.is_ok());
-
-        // Verify cache has fresh timestamp
-        let cached = loader.service_cache.read().await.get("s3").cloned();
-        assert!(cached.is_some());
-        let (_, timestamp) = cached.unwrap();
-        let elapsed = SystemTime::now().duration_since(timestamp).unwrap();
-        assert!(elapsed < Duration::from_secs(10));
+        assert_eq!(result.unwrap().unwrap().service_name, "s3");
```
mschlaipfer, line 801: Does this assert something useful? Looks like we ensured that the cache was refreshed before, but now we just check that "s3" is still in the cache.

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -720,18 +812,12 @@ mod tests {
...
-        let cache_path = RemoteServiceReferenceLoader::get_cache_path("test_fresh");
-        let _ = fs::write(&cache_path, "test content").await;
-
-        assert!(RemoteServiceReferenceLoader::is_cache_valid(&cache_path).await);
-        let _ = fs::remove_file(&cache_path).await;
+    async fn test_get_cache_meta_path() {
+        let meta_path = RemoteServiceReferenceLoader::get_cache_meta_path("s3");
+        assert!(
+            meta_path.ends_with("IAMPolicyAutopilot/s3.modified")
+                || meta_path.ends_with("IAMAutoPilot\\s3.modified")
```
mschlaipfer: Why do we allow both names, we should probably use `IAMPolicyAutopilot`?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -448,15 +516,14 @@ impl RemoteServiceReferenceLoader {
...
-    async fn is_cache_valid(path: &PathBuf) -> bool {
-        if let Ok(metadata) = fs::metadata(path).await {
-            if let Ok(modified) = metadata.modified() {
-                if let Ok(elapsed) = SystemTime::now().duration_since(modified) {
-                    return elapsed < Duration::from_secs(DEFAULT_CACHE_DURATION_IN_SECONDS);
-                }
-            }
-        }
-        false
+    fn get_cache_meta_path(service_name: &str) -> PathBuf {
```
mschlaipfer: Why `meta_path`? Maybe better would be `get_cache_modified_path`? Do we expect anything other than `.modified`? Should we then have a different file format?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -475,78 +542,100 @@ impl RemoteServiceReferenceLoader {
...
-                }
-            }
+        // Check in-memory cache (already validated by refresh_index)
+        if let Some((cached, _)) = self.service_cache.read().await.get(service_name) {
+            return Ok(Some(cached.clone()));
         }
 
-        // check temp file
-        let cache_path = Self::get_cache_path(service_name);
-        if !self.disable_file_system_cache && Self::is_cache_valid(&cache_path).await {
```
mschlaipfer, line 499: The diff here is a bit hard to follow, why did we get rid of `is_cache_valid`?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -475,78 +542,100 @@ impl RemoteServiceReferenceLoader {
...
-            if let Ok(content) = fs::read_to_string(&cache_path).await {
-                if let Ok(service_ref) = JsonProvider::parse::<ServiceReference>(&content).await {
-                    self.service_cache.write().await.insert(
-                        service_name.to_string(),
-                        (service_ref.clone(), SystemTime::now()),
-                    );
-                    return Ok(Some(service_ref));
-                }
+        // Lazily initialize the mapping if refresh_index() hasn't been called yet
+        {
```
mschlaipfer: why is this within a block?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -475,78 +542,100 @@ impl RemoteServiceReferenceLoader {
...
-                        )
-                    })?;
-                // persist content into the temp file as well
-                if !self.disable_file_system_cache {
-                    let _ = fs::write(&cache_path, &service_reference_content).await;
+        // Resolve the entry from the index
+        let mapping_guard = self.service_reference_mapping.read().await;
+        let mapping = match mapping_guard.as_ref() {
+            Some(m) => m,
+            None => return Ok(None),
```
mschlaipfer: What does the guard being `None` here mean, and why is it okay to return `Ok(None)`?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -475,78 +542,100 @@ impl RemoteServiceReferenceLoader {
         &self,
         service_name: &str,
     ) -> crate::errors::Result<Option<ServiceReference>> {
-        if let Some((cached, timestamp)) = self.service_cache.read().await.get(service_name) {
-            if let Ok(elapsed) = SystemTime::now().duration_since(*timestamp) {
-                if elapsed < Duration::from_secs(DEFAULT_CACHE_DURATION_IN_SECONDS) {
-                    return Ok(Some(cached.clone()));
-                }
-            }
+        // Check in-memory cache (already validated by refresh_index)
```
mschlaipfer: `(already validated by refresh_index)`: Why do we need `refresh_index` and `load`, and why do they share logic?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -475,78 +542,100 @@ impl RemoteServiceReferenceLoader {
...
-        if !self.disable_file_system_cache && Self::is_cache_valid(&cache_path).await {
-            if let Ok(content) = fs::read_to_string(&cache_path).await {
-                if let Ok(service_ref) = JsonProvider::parse::<ServiceReference>(&content).await {
-                    self.service_cache.write().await.insert(
-                        service_name.to_string(),
-                        (service_ref.clone(), SystemTime::now()),
-                    );
-                    return Ok(Some(service_ref));
-                }
+        // Lazily initialize the mapping if refresh_index() hasn't been called yet
```
mschlaipfer: similar comment about `refresh_index` here as above.

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -475,78 +542,100 @@ impl RemoteServiceReferenceLoader {
...
-                            format!(
-                                "Failed to parse service reference content. Detailed error: {e}"
-                            ),
-                            e,
-                        )
-                    })?;
-                // persist content into the temp file as well
-                if !self.disable_file_system_cache {
-                    let _ = fs::write(&cache_path, &service_reference_content).await;
+        // Resolve the entry from the index
```
mschlaipfer: Could we factor out some code from `load`? Right now it does a lot.

It would be nice to see immediately:
1. load from in-memory cache
2. load from filesystem cache
3. load from remote and update filesystem cache

PR #186, opened 2026-05-04
feat: add external library model support

adpaco:
*Description of changes:* IPA currently misses permissions when users call AWS services through third-party libraries like `aws_lambda_powertools`. For example, `parameters.get_parameter(...)` calls `ssm:GetParameter` internally, but since the boto3 call happens inside the library, IPA never sees it.

This PR adds external library models: JSON files that describe how library function calls map to underlying AWS SDK operations. IPA ships with a built-in model for `aws_lambda_powertools` (Python) and supports user-provided models via `--library-models <PATH>` for custom or less common libraries.

**How it works**

At startup, the `LibraryModelRegistry` loads built-in models (embedded via `rust-embed`) and optionally user-provided models from disk. During extraction, the `LibraryCallExtractor` resolves Python imports, matches function calls against loaded model patterns, and produces `SdkMethodCall` entries. Library-derived calls bypass the disambiguator (they already have known services) and merge into the pipeline before enrichment.

**Example**

```
$ iam-policy-autopilot generate-policies handler.py --region us-east-1 --account 123456789012
```

Now detects `ssm:GetParameter` and `secretsmanager:GetSecretValue` from `aws_lambda_powertools.utilities.parameters` calls.

**Scope and known limitations**

- Python only for now (schema supports all languages for forward compatibility)
- Only `ModuleLevel` call patterns are matched. `InstanceMethod` patterns (e.g., `SSMProvider().get(...)`) are deferred until variable type tracking is available
- Wildcard imports (`from X import *`) are not resolved yet
- Parameter constraints exist in the schema but no shipped model uses them; the fallback semantics need revisiting
- MCP server doesn't expose `--library-models` yet


iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
```diff
@@ -121,11 +139,25 @@ impl Extractor for PythonExtractor {
                     let paginator_calls = paginator_extractor.extract_paginate_method_calls(ast);
                     method_calls.extend(paginator_calls);
 
-                    // Clone the method calls to pass to disambiguate_method_calls
+                    // Extract external library calls separately. They already
+                    // have known `possible_services` from the model and must NOT
+                    // go through the disambiguator. The disambiguator's parameter-shape
```
mschlaipfer, line 139: Would be awesome to change the design of the pre-Java extractors eventually to match what Java does, so we don't need to differentiate between what must and what must not go through disambiguation.

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
```diff
@@ -0,0 +1,1103 @@
...
+    pub module_path: String,
+    /// Function or method name (e.g., "get_parameter")
+    pub function_name: String,
+    /// Whether this is a module-level call or instance method call
+    pub call_type: CallType,
+    /// AWS SDK operations this call maps to
+    pub sdk_operations: Vec<SdkOperationMapping>,
+    /// Optional parameter constraints for conditional matching
+    #[serde(default)]
+    pub parameter_constraints: Vec<ParameterConstraint>,
```
mschlaipfer: I don't see this used, why do we need it?

mschlaipfer:
Before I review this deeply, I'm wondering how this differs from [existing models](https://github.com/awslabs/iam-policy-autopilot/blob/a433fc0335c95e6ed7c3b00dcb8d11401bf5e8d7/iam-policy-autopilot-policy-generation/resources/config/sdks/boto3_utilities_mapping.json). It looks like this PR introduces a lot of new machinery in [external_library_models/mod.rs](https://github.com/adpaco-aws/iam-policy-autopilot/blob/ccb0de6f72b8fde96c8d16b8cb80152211c9df93/iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs), which the utilities mapping did not need (like the changes around [this comment](https://github.com/awslabs/iam-policy-autopilot/pull/186/changes#r3188044190))?

Are we convinced we can have one model format for all supported languages (also going forward)?

iam-policy-autopilot-policy-generation/src/extraction/python/extractor.rs
adpaco: I agree, but it's a large refactor that we shouldn't include in this PR. I've opened #192 to keep track of it.

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
adpaco: It's been removed. These "parameters constraints" aren't needed, but they allowed us to work with specialized call patterns for functions/methods which used default parameters. It's somewhat similar to disambiguation through arguments but falling back to the most generic call pattern.

Anyway, I already had my doubts about including it. Since it's an advanced feature that we may not need and also quite specific to this language, it's better to not include it.

adpaco:
> Before I review this deeply, I'm wondering how this differs from [existing models](https://github.com/awslabs/iam-policy-autopilot/blob/a433fc0335c95e6ed7c3b00dcb8d11401bf5e8d7/iam-policy-autopilot-policy-generation/resources/config/sdks/boto3_utilities_mapping.json). It looks like this PR introduces a lot of new machinery in [external_library_models/mod.rs](https://github.com/adpaco-aws/iam-policy-autopilot/blob/ccb0de6f72b8fde96c8d16b8cb80152211c9df93/iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs), which the utilities mapping did not need (like the changes around [this comment](https://github.com/awslabs/iam-policy-autopilot/pull/186/changes#r3188044190))?

The main difference is that external library models handle code that doesn't use the SDK at all. In contrast, the utilities mapping describes alternative ways to call the SDK, but are still SDK calls. For example, `parameters.get_parameter(...)` isn't recognizable by the existing `$OBJ.$METHOD($$$ARGS)` pattern matching because it's not a boto3 call. We need import resolution to match general calls against our loaded models.

The `filter_map` handling is needed because library-derived calls already have their service resolved (from the model) and can't go through disambiguation. (Opened #187 to track unifying this with the Java approach long-term.)

> Are we convinced we can have one model format for all supported languages (also going forward)?

The schema introduced in this PR maps to any language with imports and function/method calls. What's language-specific is the extractor that interprets it (import resolution, call matching), not the model format. If a future language needs something the current schema can't express, we'd extend it with an optional field rather than fork the format.

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+
+            let embedded_file = if let Some(f) = ExternalLibraryModelsAsset::get(file_name.as_ref())
+            {
+                f
+            } else {
+                log::warn!(
+                    "Failed to read embedded external library model '{}'",
+                    file_path.display()
+                );
+                continue;
```
mschlaipfer: Wouldn't it make sense to fail here?

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        Ok(Self { models })
+    }
+
+    /// Get all models for a given language.
+    pub(crate) fn models_for_language(&self, language: Language) -> Vec<&ExternalLibraryModel> {
+        self.models
+            .values()
+            .filter(|model| model.language == language)
+            .collect()
```
mschlaipfer: Should self.models be a HashMap<Language, Vec<ExternalLibraryModel>> so we can use .get(...)? Right now it seems that we don't use .get(...) anywhere, which suggests that the way we store the data isn't right.

Maybe `ExternalLibraryModel` should store the library name?

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// AWS SDK operations this call maps to
+    pub sdk_operations: Vec<SdkOperationMapping>,
+}
+
+/// Discriminator for how the function is invoked.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum CallType {
+    /// Called on an imported module: `parameters.get_parameter(...)`
+    ModuleLevel,
```
mschlaipfer: What about static methods? Is this the right name?

iam-policy-autopilot-policy-generation/resources/config/external-library-models/aws_lambda_powertools.json
```diff
@@ -0,0 +1,29 @@
...
+      "call_type": "module_level",
+      "sdk_operations": [
+        {
+          "service": "secretsmanager",
+          "operation": "GetSecretValue"
+        }
+      ]
+    }
+  ]
+}
```
mschlaipfer, line 29: Seems like we're all over the place with casing in utility models, so probably fine to use snake_case here.

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub operation: String,
+}
+
+/// Registry of external library models, indexed by (language, library_name).
+///
+/// The registry discovers, loads, and indexes all built-in external library
+/// models for a given language. Built-in models are embedded at compile time
+/// via `rust-embed`.
+pub(crate) struct LibraryModelRegistry {
+    /// Models indexed by (language, library_name) for fast lookup.
```
mschlaipfer: It says fast lookup here, but I think the only lookup is via iteration over `.values()`

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+#[folder = "resources/config/external-library-models"]
+#[include = "*.json"]
+struct ExternalLibraryModelsAsset;
+
+/// Represents a deserialized external library model file.
+///
+/// Each model describes one library for one language, mapping its function calls
+/// to underlying AWS SDK operations.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub(crate) struct ExternalLibraryModel {
```
mschlaipfer, line 23: Since we're not enabling to "bring your own library model" as of now, I'm fine with this. I think otherwise this would need to be more flexible, as this format would be a one-way door.

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// Library name (e.g., "aws_lambda_powertools")
+    pub library_name: String,
+    /// Target programming language
+    pub language: Language,
+    /// Semver version constraint (informational, not enforced at runtime)
+    #[serde(default)]
+    pub version: Option<String>,
```
mschlaipfer, line 30: What version is this? The version of the model, I assume, and not the version of the schema? Could you clarify the comment?

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    ///
+    /// Iterates over all `.json` files in the embedded `external-library-models/`
+    /// directory, parses each one, and collects models that match
+    /// the requested language. If a built-in model fails to parse, a warning
+    /// is logged and the model is skipped (extraction is not halted).
+    fn load_builtin_models(language: Language) -> Vec<ExternalLibraryModel> {
+        let mut models = Vec::new();
+
+        for file_name in ExternalLibraryModelsAsset::iter() {
+            let file_path = std::path::Path::new(file_name.as_ref());
```
mschlaipfer: Why does this need to be a `Path`?

iam-policy-autopilot-policy-generation/resources/config/external-library-models/aws_lambda_powertools.json
mschlaipfer, line 1: Is this complete?

iam-policy-autopilot-cli/tests/integration_tests.rs
```diff
@@ -861,3 +861,173 @@ fn test_dictionary_unpacking_file() {
...
+        all_actions.iter().any(|a| a == "ssm:GetParameter"),
+        "Policy should include ssm:GetParameter action. Actions: {all_actions:?}"
+    );
+    assert!(
+        all_actions
+            .iter()
+            .any(|a| a == "secretsmanager:GetSecretValue"),
+        "Policy should include secretsmanager:GetSecretValue action. Actions: {all_actions:?}"
+    );
+}
```
mschlaipfer, line 1059: Not for this PR: This file is getting large, I noticed agents struggling with files starting around the 1kLOC mark.

iam-policy-autopilot-policy-generation/src/extraction/python/library_call_extractor.rs
```diff
@@ -0,0 +1,894 @@
...
+            if child_text.as_ref() == ","
+                || child_text.as_ref() == "("
+                || child_text.as_ref() == ")"
+            {
+                continue;
+            }
+
+            if child_kind == import_node_kinds::WILDCARD_IMPORT {
+                // TODO: wildcard imports (`from X import *`) could be resolved
+                // using the loaded model's function names for the module path.
```
mschlaipfer: Can we address this? How do we handle this so far for boto3?

iam-policy-autopilot-policy-generation/src/extraction/python/library_call_extractor.rs
```diff
@@ -0,0 +1,894 @@
...
+        resolved_imports: &HashMap<String, String>,
+        pattern_module_path: &str,
+    ) -> bool {
+        if let Some(canonical_path) = resolved_imports.get(object_text) {
+            // The canonical path should match the pattern's module_path exactly,
+            // or the canonical path should be a prefix that the pattern's module_path
+            // starts with (for cases like `import X.Y` where local name is `X`
+            // and canonical is `X.Y`, matching pattern module_path `X.Y`).
+            canonical_path == pattern_module_path
+                || pattern_module_path.starts_with(&format!("{canonical_path}."))
```
mschlaipfer: I'm not sure I fully understand why we need `starts_with`.

iam-policy-autopilot-policy-generation/src/extraction/python/library_call_extractor.rs
```diff
@@ -0,0 +1,894 @@
...
+        assert_eq!(imports.len(), 2);
+
+        let canonical = imports.get("parameters").expect("should have 'parameters'");
+        assert_eq!(canonical, "aws_lambda_powertools.utilities.parameters");
+
+        let canonical = imports
+            .get("idempotency")
+            .expect("should have 'idempotency'");
+        assert_eq!(canonical, "aws_lambda_powertools.utilities.idempotency");
+    }
```
mschlaipfer, line 643: Could these be `rstest`s?

iam-policy-autopilot-policy-generation/src/extraction/python/library_call_extractor.rs
```diff
@@ -0,0 +1,894 @@
...
+                    metadata.location.start_position.0
+                );
+                // The call is on line 2 (1-based)
+                prop_assert_eq!(
+                    metadata.location.start_position.0,
+                    2,
+                    "SdkMethodCall[{}] metadata line should be 2 (call is on second line) but got {}",
+                    i,
+                    metadata.location.start_position.0
+                );
```
mschlaipfer, line 888: This assertion implies the one above, but I don't understand how every call is at line 2 (since this is in a for loop). Is there at most 1 result? It seems like `arb_sdk_operations` has `1..=3`, so I'm confused.

iam-policy-autopilot-policy-generation/src/extraction/python/library_call_extractor.rs
```diff
@@ -0,0 +1,894 @@
...
+                        service,
+                        operation,
+                    }
+                },
+            ),
+            1..=3,
+        )
+    }
+
+    proptest! {
```
mschlaipfer, line 767: Nice to add a prop test!

iam-policy-autopilot-policy-generation/src/extraction/python/library_call_extractor.rs
```diff
@@ -0,0 +1,894 @@
...
+    // ---- Unit tests for library call extraction ----
+
+    /// Helper: load the built-in LibraryModelRegistry for Python.
+    fn load_python_registry() -> LibraryModelRegistry {
+        LibraryModelRegistry::load(crate::Language::Python)
+            .expect("built-in Python registry should load successfully")
+    }
+
+    #[test]
+    fn test_get_parameter_with_from_import_produces_ssm_sdk_call() {
```
mschlaipfer: Can these be `rstest`s?

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
adpaco: Yes, we are now doing as with other embedded data (panic).

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
adpaco: Sorry for this oversight. The key type I was using initially didn't end up matching how we access the models in practice. I have updated this to use the proposed type.

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
adpaco: Good point. After some consideration, I've updated the `CallType` enum to be as follows:
```rust
pub(crate) enum CallType {
    /// A function defined in a module/namespace: `mod.func()`, `pkg.Func()`, `func()`
    Function,
    /// A method called on a class/type: `Class.method()`, `Type::method()`
    StaticMethod,
    /// A method called on an instance: `obj.method()`, `obj->method()`
    InstanceMethod,
}
```

I think this is a good design for two reasons:
 1. It can be used for all languages we currently support and is a good fit for other languages like Rust, C/C++ or PHP.
 2. It's using very common terms for function definition types so it's understood by everyone.

Please let me know if you have any suggestions about these names.

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
adpaco: The `Hashmap` type has changed and they're indexed by language only.

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
adpaco: It's the version of the library we built the model with. Updated the comment.

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
adpaco: Removed.

iam-policy-autopilot-policy-generation/resources/config/external-library-models/aws_lambda_powertools.json
adpaco: No. We could add a few more entries if you like but I'd prefer for these models to be completed using IPA itself.

iam-policy-autopilot-policy-generation/src/extraction/python/library_call_extractor.rs
adpaco: Thanks, I wrote down some cases and realized I was wrong - we don't need `starts_with` here.

iam-policy-autopilot-policy-generation/src/extraction/python/library_call_extractor.rs
adpaco: Done.

iam-policy-autopilot-policy-generation/src/extraction/python/library_call_extractor.rs
adpaco: All tests are `rstest`s now.

iam-policy-autopilot-policy-generation/src/extraction/python/library_call_extractor.rs
adpaco: Removed the assertion. Added a comment explaining why all results come from line 2 and the call may produce 1-3 operations.

iam-policy-autopilot-policy-generation/src/extraction/python/library_call_extractor.rs
adpaco: Ok, I added support for wildcard imports. Let me explain a little because it's not a trivial feature.

We don't need to handle wildcard imports for boto3 because the boto3 extractor doesn't look at imports. It directly matches `$OBJ.$METHOD($$$ARGS)` calls and uses disambiguation by matching parameters to determine if they're SDK calls.

This library call extractor, however, cannot match function/method calls with library call patterns the same way, and import resolution is the primary method to determine that instead (using module paths). Wildcard imports in practice bring function/method declarations into the module as if they were defined there. There are different ways we could handle them, but in the end I went with this option:
1. We record wildcard imports separately
2. Before matching, we expand the imports using the model's call pattern. In other words, we add one import entry per function/method declaration in the wildcard-imported module.
3. We match call patterns as usual, no need for special matching logic

The other challenge that wildcard import introduces is that now we need to match with the more generic `$FUNC($$$ARGS)`. I rewrote the `ast-grep`ing part to keep doing it in one pass, routing to the appropriate case depending on the node type.

Of course, this would probably be much simpler if we had fully qualified name resolution from LSP integration . We do need it anyway for the `InstanceMethod` call type.

mschlaipfer (dismissed):

iam-policy-autopilot-cli/tests/integration_tests.rs
```diff
@@ -861,3 +861,173 @@ fn test_dictionary_unpacking_file() {
...
+            );
+            assert!(
+                statement.get("Action").is_some(),
+                "Statement should have Action"
+            );
+            assert!(
+                statement.get("Resource").is_some(),
+                "Statement should have Resource"
+            );
+
```
janstey-amzn, line 1010: Do we want to go a little deeper here if we know the actions that should be allowed, and confirm the resource ARN(s) contain the account Id and reigon?

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    fn load_builtin_models(language: Language) -> Vec<ExternalLibraryModel> {
+        let mut models = Vec::new();
+
+        for file_name in ExternalLibraryModelsAsset::iter() {
+            let embedded_file =
+                ExternalLibraryModelsAsset::get(file_name.as_ref()).unwrap_or_else(|| {
+                    panic!(
+                        "Failed to read embedded external library model '{}'",
+                        file_name.as_ref()
+                    )
```
janstey-amzn, line 97: The description of this function points out that this function fails gracefully, in this event of a failed file read, should we gracefully skip as well?

janstey-amzn (dismissed):

iam-policy-autopilot-cli/tests/integration_tests.rs
adpaco: Sounds good, added assertions to check the account ID and region passed via CLI flags.

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
adpaco: We updated this after an earlier review that suggested we should fail for file read. That is consistent with how other embedded data is handled in the codebase. I've updated the doc comments to clarify we:
 * Panic for file reads (should be unreachable)
 * Warn+skip for JSON parsing failures.

iam-policy-autopilot-policy-generation/src/extraction/external_library_models/mod.rs
```diff
@@ -90,16 +90,7 @@ impl LibraryModelRegistry {
 
         let mut models: HashMap<Language, Vec<ExternalLibraryModel>> = HashMap::new();
         for model in builtin_models {
```
mschlaipfer: We can get rid of this loop now right?

mschlaipfer (approved):

janstey-amzn (approved):

PR #188, opened 2026-05-04
fix(enrichment): include source in Operation equality so --explain keeps every call site (#183)

### LINKED ISSUE #183: `--explain` drops duplicate call sites when the same SDK method appears multiple times
## Reproduction

```python
import boto3
s3 = boto3.client("s3")

def upload_avatar(user_id, data):
    s3.put_object(Bucket="avatars-bucket", Key=f"avatars/{user_id}.png", Body=data)

def upload_document(doc_id, content):
    s3.put_object(Bucket="documents-bucket", Key=f"docs/{doc_id}.json", Body=content)
```

```bash
iam-policy-autopilot generate-policies test.py --explain "s3:*" --pretty
```

## Expected

Each call site should appear as a separate **Reason** (the `Operations` array within a Reason is a provenance chain, e.g. `[source_op, fas_op]`, not a list of call sites):

```json
"s3:PutObject": [
  {
    "Operations": [{ "Service": "s3", "Name": "PutObject",
        "Source": { "Expr": "s3.put_object(Bucket=\"avatars-bucket\", ...)", "Location": "test.py:5.5-5.78" }}]
  },
  {
    "Operations": [{ "Service": "s3", "Name": "PutObject",
        "Source": { "Expr": "s3.put_object(Bucket=\"documents-bucket\", ...)", "Location": "test.py:8.5-8.82" }}]
  }
]
```

## Actual

Only the first call site is kept; the second is silently dropped:

```json
"s3:PutObject": [
  {
    "Operations": [{ "Service": "s3", "Name": "PutObject",
        "Source": { "Expr": "s3.put_object(Bucket=\"avatars-bucket\", ...)", "Location": "test.py:5.5-5.78" }}]
  }
]
```

## Root Cause

The custom `PartialEq`/`Hash` on `Operation` compares only `service`, `name`, and `context()` — it **ignores the `source` field** (which holds the file location). Since `Reason` derives `PartialEq`/`Hash` from `Vec<Arc<Operation>>`, two Reasons pointing to different source locations are considered identical.

`Explanation::merge()` deduplicates Reasons via a `HashSet`, so the second call site's Reason is treated as a duplicate and dropped.

## Fix

Include `source` in `Operation`'s `PartialEq` and `Hash` implementations. This equality is only consumed by `Explanation::merge()` (via `Reason`'s derived traits) — it is not used in the policy generation dedup path, so changing it won't affect policy output.

The existing test `test_explanation_deduplication` uses `OperationSource::Provided` (no metadata) and would need a companion test with `OperationSource::Extracted` carrying distinct locations.

ChrisJr404:
Closes #183.

The custom \`PartialEq\`/\`Hash\` on \`Operation\` compared \`service\` + \`name\` + \`context()\` but ignored the \`source\` field that holds the file location for extracted operations. \`Reason\` derives \`PartialEq\`/\`Hash\` from \`Vec<Arc<Operation>>\`, and \`Explanation::merge\` dedupes \`Reason\`s through a \`HashSet\` — so two reasons pointing at different call sites of the same SDK method collapsed into one and the second was silently dropped from \`--explain\` output.

The repro from the issue (two \`s3.put_object\` calls into different buckets) yielded one Reason instead of two; this PR makes both survive.

Add \`source\` to both \`PartialEq\` and \`Hash\`. The existing inline comment already noted the equality was "directly relevant to FAS expansion" — and that path stays correct here because FAS context lives inside \`source\` via \`OperationSource::Fas\`. The policy-generation dedup path keys on \`service:name\` rather than on \`Operation\` equality, so emitted policies are unaffected.

## Tests

- Existing \`test_explanation_deduplication\` still passes (uses \`OperationSource::Provided\` on both sides; \`Provided == Provided\` regardless of the new field, so the equivalent-Reason dedup path is preserved).
- New \`test_explanation_preserves_distinct_call_sites\` is the companion you flagged in the issue: two \`OperationSource::Extracted\` reasons with distinct \`Location\`s, asserts both survive merge dedup. This is the scenario that was previously dropping a Reason on the floor.
- Existing \`test_explanation_with_fas_expansion\` and \`test_explanation_with_possible_false_positive\` still pass.

\`\`\`
$ cargo test -p iam-policy-autopilot-policy-generation test_explanation
test policy_generation::engine::tests::test_explanation_with_fas_expansion ... ok
test policy_generation::engine::tests::test_explanation_preserves_distinct_call_sites ... ok
test policy_generation::engine::tests::test_explanation_deduplication ... ok
test policy_generation::engine::tests::test_explanation_with_possible_false_positive ... ok
\`\`\`

## Test plan

- [x] All 4 explanation-related tests green
- [x] Build clean (\`cargo build -p iam-policy-autopilot-policy-generation\`)
- [ ] (For maintainer) — run the issue's exact repro (\`generate-policies test.py --explain "s3:*" --pretty\` against the avatars/documents put_object snippet) and confirm both call sites now appear under \`s3:PutObject\`.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
@@ -7,7 +7,7 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use super::{Action, Context, EnrichedSdkMethodCall, Explanation, Reason, Resource};
+use super::{Action, Context, EnrichedSdkMethodCall, Explanation, OperationKey, Reason, Resource};
 use crate::enrichment::operation_fas_map::{OperationFasMap, OperationFasMaps};
 use crate::enrichment::service_reference::ServiceReference;
 use crate::enrichment::{Condition, Operation, ServiceReferenceLoader};
@@ -15,9 +15,20 @@ use crate::errors::{ExtractorError, Result};
 use crate::service_configuration::ServiceConfiguration;
 use crate::{SdkMethodCall, SdkType};
 
+/// A node in the FAS expansion dependency graph.
+#[derive(Clone, Debug)]
+struct FasNode {
+    /// The actual operation this node represents.
+    operation: Arc<Operation>,
+    /// Parent operations that triggered this one in the FAS expansion.
+    parents: Vec<OperationKey>,
+}
+
+/// FAS expansion result, keyed by `OperationKey` (service + name + context)
+/// so that cycle detection ignores `source` information.
 #[derive(Clone, Debug)]
 struct FasExpansion {
-    dependency_graph: HashMap<Arc<Operation>, Vec<Arc<Operation>>>,
+    nodes: HashMap<OperationKey, FasNode>,
 }
 
 impl FasExpansion {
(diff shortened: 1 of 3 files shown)


mschlaipfer:
Hey @ChrisJr404, thanks for the contribution!

It turns out that my analysis on the issue wasn't right (there are some failing unit tests). I had missed the part where we need a different type of equality for FAS expansion and for Explanation merging. The right fix needs a refactoring of FAS expansion. I am going to update this PR, since I already made the change as part of my investigation. Hope that's fine with you!

mschlaipfer (approved):

mschlaipfer:
Also, we require signed commits (https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), if you could please address that as well.

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -7,17 +7,28 @@
...
 use crate::service_configuration::ServiceConfiguration;
 use crate::{SdkMethodCall, SdkType};
 
+/// A node in the FAS expansion dependency graph.
+#[derive(Clone, Debug)]
+struct FasNode {
+    /// The actual operation this node represents.
+    operation: Arc<Operation>,
+    /// Operations that depend on (triggered) this one.
+    dependencies: Vec<OperationKey>,
```
adpaco: Nit: It's more like these are the parents. I'd rename this to `parents` or `triggers` to make it clearer.

adpaco-aws (approved):

mschlaipfer (approved):

adpaco (approved):

PR #190, opened 2026-05-05
feat: support namespace imports in TypeScript/JavaScript

### LINKED ISSUE #95: Support for JavaScript/TypeScript Wildcard Imports in AWS SDK Extraction
**Is your feature request related to a problem? Please describe.**

I'm unable to generate policies for JavaScript/TypeScript codebases that use wildcard import syntax like `import * as S3 from '@aws-sdk/client-s3'`. The current extractor only detects AWS SDK operations when using explicit named imports or destructuring imports, but many codebases use namespace imports for better code organization. This results in incomplete IAM policy generation because waiter calls, command instantiations, and paginator usage are not detected when accessed through a namespace (e.g., `S3.waitUntilBucketExists()`, `new S3.CreateBucketCommand()`).

**Describe the solution you'd like**

I'd like the JavaScript/TypeScript extractor to support wildcard import patterns and detect AWS SDK operations accessed through namespaces. Specifically:

1. **Import Detection**: Recognize `import * as NameSpace from '@aws-sdk/client-service'` patterns
2. **Operation Detection**: Detect namespace-prefixed calls like:
   - `NameSpace.waitUntilBucketExists(config, params)` (waiters)
   - `new NameSpace.CreateBucketCommand(params)` (commands)  
   - `NameSpace.paginateListObjects(config, params)` (paginators)
3. **Service Resolution**: Map the namespace back to the correct AWS service for policy generation

**Additional context**

**Current Working Pattern:**
```javascript
// ✅ Currently supported
import { S3Client, waitUntilBucketExists, CreateBucketCommand } from '@aws-sdk/client-s3';

await waitUntilBucketExists({ client }, { Bucket: 'my-bucket' });
const command = new CreateBucketCommand({ Bucket: 'my-bucket' });
```

**Requested Pattern:**
```javascript
// ❌ Not currently supported
import * as S3 from '@aws-sdk/client-s3';

await S3.waitUntilBucketExists({ client }, { Bucket: 'my-bucket' });
const command = new S3.CreateBucketCommand({ Bucket: 'my-bucket' });
```

adpaco:
*Issue #, if available:* Resolves #95

*Description of changes:*

Adds support for wildcard/namespace import patterns (`import * as S3 from '@aws-sdk/client-s3'`) in the JavaScript/TypeScript extractor. Operations accessed through the namespace prefix are now detected:

- Commands: `new S3.CreateBucketCommand({ Bucket: "my-bucket" })`
- Paginators: `S3.paginateListObjectsV2({ client }, { Bucket: "my-bucket" })`
- Waiters: `await S3.waitUntilBucketExists({ client }, { Bucket: "my-bucket" })`
- Library classes: `new S3Storage.Upload({ client, params })`

Implementation:
- Added `wildcard_namespace` field to `SublibraryInfo` to track namespace aliases
- Extended `scan_aws_imports` to detect `import * as NS from '@aws-sdk/...'` patterns
- Integrated wildcard scanning into existing extractors (`extract_command_operations`, `extract_paginate_operations`, `extract_waiter_operations`, `extract_library_class_operations`)
- Added generic `find_namespace_calls` scanner method using `$$$ARGS` for variadic matching
- Includes lib-* expansion support for the rare case of wildcard-imported library packages

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/extraction/javascript/extractor.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/javascript/extractor.rs
@@ -524,4 +524,212 @@ const command = new GetObjectCommand({ Bucket: 'test', Key: 'test.txt' });
             _ => panic!("Should return JavaScript result"),
         }
     }
+
+    #[tokio::test]
+    async fn test_wildcard_import_command_extraction() {
+        let extractor = JavaScriptExtractor::new();
+        let source_code = r#"
+import * as S3 from "@aws-sdk/client-s3";
+
+const client = new S3.S3Client({ region: "us-east-1" });
+
+async function createBucket() {
+    const command = new S3.CreateBucketCommand({ Bucket: "my-bucket" });
+    await client.send(command);
+}
+        "#;
+
+        let result = extractor.parse(&create_source_file(source_code)).await;
+        let method_calls = result.method_calls_ref();
+
+        // Should find CreateBucket operation from namespace-prefixed command
+        assert!(
+            !method_calls.is_empty(),
+            "Should find operations from wildcard import"
+        );
+
+        let create_bucket_op = method_calls
+            .iter()
+            .find(|call| call.name == "CreateBucket")
+            .expect("Should find CreateBucket operation from S3.CreateBucketCommand");
+
+        // Should be associated with s3 service
+        assert_eq!(
+            create_bucket_op.possible_services,
+            vec!["s3"],
+            "Should associate with s3 service"
+        );
+
+        println!("✅ Wildcard import command extraction works!");
+    }
+
+    #[tokio::test]
+    async fn test_wildcard_import_waiter_extraction() {
+        let extractor = JavaScriptExtractor::new();
+        let source_code = r#"
+import * as S3 from "@aws-sdk/client-s3";
+
+const client = new S3.S3Client({ region: "us-east-1" });
+
+async function waitForBucket() {
+    await S3.waitUntilBucketExists(
+        { client, maxWaitTime: 300 },
+        { Bucket: "my-bucket" }
+    );
+}
+        "#;
+
+        let result = extractor.parse(&create_source_file(source_code)).await;
+        let method_calls = result.method_calls_ref();
+
+        // Should find waiter operation from namespace-prefixed waiter
+        assert!(
+            !method_calls.is_empty(),
+            "Should find waiter from wildcard import"
+        );
+
+        // The waiter name "BucketExists" will be resolved to actual operation in filter_map
+        let waiter_call = method_calls
+            .iter()
+            .find(|call| call.name == "BucketExists")
+            .expect("Should find BucketExists waiter from S3.waitUntilBucketExists");
+
+        // Should be associated with s3 service
+        assert_eq!(
+            waiter_call.possible_services,
+            vec!["s3"],
+            "Should associate with s3 service"
+        );
+
+        println!("✅ Wildcard import waiter extraction works!");
+    }
+
+    #[tokio::test]
+    async fn test_wildcard_import_paginator_extraction() {
+        let extractor = JavaScriptExtractor::new();
+        let source_code = r#"
+import * as S3 from "@aws-sdk/client-s3";
+
+const client = new S3.S3Client({ region: "us-east-1" });
+
+async function listAllObjects() {
+    const paginator = S3.paginateListObjectsV2(
+        { client },
+        { Bucket: "my-bucket" }
+    );
+    
+    for await (const page of paginator) {
+        console.log(page.Contents);
+    }
+}
+        "#;
+
+        let result = extractor.parse(&create_source_file(source_code)).await;
+        let method_calls = result.method_calls_ref();
+
+        // Should find paginator operation from namespace-prefixed paginator
+        assert!(
+            !method_calls.is_empty(),
+            "Should find paginator from wildcard import"
+        );
+
+        let list_op = method_calls
+            .iter()
+            .find(|call| call.name == "ListObjectsV2")
+            .expect("Should find ListObjectsV2 operation from S3.paginateListObjectsV2");
+
+        // Should be associated with s3 service
+        assert_eq!(
+            list_op.possible_services,
+            vec!["s3"],
+            "Should associate with s3 service"
+        );
+
+        println!("✅ Wildcard import paginator extraction works!");
+    }
+
+    #[tokio::test]
+    async fn test_wildcard_import_multiple_services() {
+        let extractor = JavaScriptExtractor::new();
+        let source_code = r#"
+import * as S3 from "@aws-sdk/client-s3";
+import * as DynamoDB from "@aws-sdk/client-dynamodb";
+
+async function multiServiceOperations() {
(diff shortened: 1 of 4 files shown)


iam-policy-autopilot-policy-generation/src/extraction/javascript/extractor.rs
```diff
@@ -524,4 +524,212 @@ const command = new GetObjectCommand({ Bucket: 'test', Key: 'test.txt' });
...
+
+        assert_eq!(put_item_op.possible_services, vec!["dynamodb"]);
+
+        // Verify no double-extraction: should appear exactly once
+        let put_item_count = method_calls
+            .iter()
+            .filter(|call| call.name == "PutItem")
+            .count();
+        assert_eq!(put_item_count, 1, "PutItem should not be double-extracted");
+    }
```
mschlaipfer, line 628: Could we make these tests declarative or introduce test fixtures, if we don't have them at this point yet? Hard to review these imperative tests which all assert something different.

adpaco: Yeah, no problem! I have converted these into `rstest` cases.

iam-policy-autopilot-policy-generation/src/extraction/javascript/extractor.rs
```diff
@@ -524,4 +524,101 @@ const command = new GetObjectCommand({ Bucket: 'test', Key: 'test.txt' });
...
+        r#"
+import * as dynamoLib from "@aws-sdk/lib-dynamodb";
+import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
+const client = dynamoLib.DynamoDBDocumentClient.from(new DynamoDBClient({}));
+async function putItem() {
+    const command = new dynamoLib.PutCommand({ TableName: "my-table", Item: { id: "1" } });
+    await client.send(command);
+}
+        "#,
+        &[("dynamodb", "PutItem")],
```
mschlaipfer, line 594: Why is this `PutItem` and not `Put`?

iam-policy-autopilot-policy-generation/src/extraction/javascript/extractor.rs
```diff
@@ -524,4 +524,101 @@ const command = new GetObjectCommand({ Bucket: 'test', Key: 'test.txt' });
...
+    #[tokio::test]
+    async fn test_wildcard_import_extraction(
+        #[case] source_code: &str,
+        #[case] expected_operations: &[(&str, &str)],
+    ) {
+        let extractor = JavaScriptExtractor::new();
+        let result = extractor.parse(&create_source_file(source_code)).await;
+        let method_calls = result.method_calls_ref();
+
+        for &(expected_service, expected_op) in expected_operations {
```
mschlaipfer, line 610: This test succeeds if you don't provide any expected operations (which might get cheated by an LLM).

iam-policy-autopilot-policy-generation/src/extraction/javascript/shared.rs
```diff
@@ -599,6 +602,54 @@ impl ExtractionUtils {
...
+                    log::debug!("No mapping found for {service}/{name}, using original name");
+                    vec![name.to_string()]
+                })
+        } else {
+            vec![name.to_string()]
+        }
+    }
+
+    /// Strip paginator/command prefixes to get the operation name.
+    fn strip_paginator_prefix(name: &str) -> String {
```
mschlaipfer: Misleading name, this also strips the suffix, or leaves it unchanged.

iam-policy-autopilot-policy-generation/src/extraction/javascript/shared.rs
```diff
@@ -475,18 +457,39 @@ impl ExtractionUtils {
...
+
+                // Wildcard imports (e.g., import * as S3Storage from '@aws-sdk/lib-storage')
+                // Handles namespace-prefixed classes like S3Storage.Upload(...)
+                if let Some(namespace) = &sublibrary_info.wildcard_namespace {
+                    let service_mappings = lib_mappings.library_operation_expansions.get(&service);
+
+                    if let Some(service_mappings) = service_mappings {
+                        for (class_name, expanded_commands) in service_mappings {
+                            // Skip classes already handled by command/paginate/waiter extractors
+                            if Self::is_command_name_pattern(class_name) {
```
mschlaipfer: This is brittle.

iam-policy-autopilot-policy-generation/src/extraction/javascript/extractor.rs
adpaco: `PutCommand` is a wrapper from `@aws-sdk/lib-dynamodb` that maps to `PutItemCommand` using the expansion mapping in [js_v3_libraries.json](https://github.com/awslabs/iam-policy-autopilot/blob/main/iam-policy-autopilot-policy-generation/resources/config/sdks/js_v3_libraries.json#L3). So `PutItem` is the actual operation name.

iam-policy-autopilot-policy-generation/src/extraction/javascript/extractor.rs
adpaco: Added an assertion to defend against empty `expected_operations`.

iam-policy-autopilot-policy-generation/src/extraction/javascript/shared.rs
adpaco: I was going to rename this to `extract_operation_name` but decided to inline since it's simple enough.

iam-policy-autopilot-policy-generation/src/extraction/javascript/shared.rs
adpaco: Updated the extraction method with a `handled_names` set that keeps track of operation names already handled elsewhere. We use that instead of `is_command_name_pattern` for skipping now.

iam-policy-autopilot-policy-generation/src/extraction/javascript/shared.rs
```diff
@@ -102,6 +104,7 @@ impl ExtractionUtils {
         T: ast_grep_language::LanguageExt,
     {
         let mut method_calls = Vec::new();
+        let mut handled_names = HashSet::new();
```
mschlaipfer, line 107: Is this per-file? Does it need to be reset anywhere? We should check how we handle this right now. It seems like we'd be missing `SdkMethodCall`s. This might matter once we include arguments into the analysis. Also for `--explain`.

Is this needed because we start from the import and find the method calls later?

adpaco: > Is this per-file? Does it need to be reset anywhere?

Yes, `extract_operations_from_imports` is called once per file and `handled_names` is created for each file. We use it to avoid re-processing names handled in earlier extractors, we don't need to reset it anywhere.

> Is this needed because we start from the import and find the method calls later?

No, it's needed to prevent double extraction between import extractors. Without it, `extract_library_class_operations` would iterate all mapping entries and try to extract calls like `PutCommand` again, even though `extract_command_operations` should have already handled it. Previously, we were using `is_command_name_pattern` to "guess" if the name had already been handled.

mschlaipfer (approved):

janstey-amzn (approved):

PR #198, opened 2026-05-12
test: add integration test projects and runner

mschlaipfer:
*Issue #, if available:*

*Description of changes:*

Adds integration tests run in a nightly GH worfklow for all 4 currently supported languages across 10 projects. For each project, the integration test runner:
1. Deploys the project's CDK stack via a deployment role (using a deployment policy provided as part of the project)
1. Creates application policies via IPA
1. Runs the application via role using the generated policy
1. Destroys the stack

The applications are LLM-generated so that they are "equal" across the 4 languages and fail early to highlight permission issues.

Successful demo run in: https://github.com/mschlaipfer/iam-policy-autopilot/actions/runs/25740045775

Future work: parallelize runs along the project dimension (not the language dimension, in order not to interfere with other executions of the application).


integration-tests/projects/run_001/cdk/lib/stack.ts
```diff
@@ -0,0 +1,211 @@
...
+        sid: 'AWSCloudTrailWrite',
+        effect: iam.Effect.ALLOW,
+        principals: [new iam.ServicePrincipal('cloudtrail.amazonaws.com')],
+        actions: ['s3:PutObject'],
+        resources: [`${this.bucket.bucketArn}/AWSLogs/*`],
+        conditions: {
+          StringEquals: { 's3:x-amz-acl': 'bucket-owner-full-control' },
+        },
+      }),
+    );
```
janstey-amzn, line 66: We're using a policy condition to enforce the bucket-owner-full-control ACL. AWS now considers S3 ACLs a legacy feature and recommends a more modern approach.

Instead of managing permissions via ACL headers, we should switch the bucket setting to S3 Object Ownership: Bucket owner enforced. This automatically gives the bucket owner full control over all uploaded objects and allows us to disable ACLs entirely, simplifying our security model and following current AWS best practices.

```
const bucket = new s3.Bucket(this, 'MyBucket', {
  objectOwnership: s3.ObjectOwnership.BUCKET_OWNER_ENFORCED,
  blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
});

const cloudtrailPrincipal = new iam.ServicePrincipal('cloudtrail.amazonaws.com')
  .withConditions({
    StringEquals: { 'aws:SourceArn': `arn:aws:cloudtrail:${this.region}:${this.account}:trail/<trail-name>` }
  });

this.bucket.grantWrite(cloudtrailPrincipal, 'AWSLogs/*');

// CloudTrail requires this to verify bucket permissions during setup
this.bucket.grant(cloudtrailPrincipal, 's3:GetBucketAcl');
```

mschlaipfer: These are LLM-generated test projects. While I generally agree, I'm not sure it's worth addressing all these warts.

integration-tests/projects/run_001/cdk/bin/app.ts
```diff
@@ -0,0 +1,39 @@
...
+ *   npx cdk destroy
+ */
+
+import 'source-map-support/register';
+import * as cdk from 'aws-cdk-lib';
+import { SecurityAnalyticsStack } from '../lib/stack';
+
+const app = new cdk.App();
+
+new SecurityAnalyticsStack(app, 'SecurityAnalyticsStack-run001-0aa559b7', {
```
janstey-amzn, line 31: I'm curious about the hash in the stack Id, is there a reason for this?

integration-tests/projects/run_003/cdk/bin/app.ts
```diff
@@ -0,0 +1,39 @@
...
+ *   npx cdk destroy
+ */
+
+import 'source-map-support/register';
+import * as cdk from 'aws-cdk-lib';
+import { DataPipelineStack } from '../lib/stack';
+
+const app = new cdk.App();
+
+new DataPipelineStack(app, 'DataPipelineStack-run002-7beb16a2', {
```
janstey-amzn, line 31: This should be `-run003-`

The use of inaccurate test run IDs repeats for all subsequent `app.ts` and `stack.ts` files, I won't add additional comments on each instance

integration-tests/projects/run_002/cdk/bin/app.ts
```diff
@@ -0,0 +1,39 @@
...
+ *   npx cdk destroy
+ */
+
+import 'source-map-support/register';
+import * as cdk from 'aws-cdk-lib';
+import { FileMonitoringStack } from '../lib/stack';
+
+const app = new cdk.App();
+
+new FileMonitoringStack(app, 'FileMonitoringStack-run001', {
```
janstey-amzn, line 31: This should be `-run002`

integration-tests/runner/src/main.rs
```diff
@@ -0,0 +1,854 @@
...
+                lang_cfg,
+                &caller_arn,
+            )
+            .await;
+            if result.success {
+                any_success = true;
+            }
+            report.language_results.insert(lang.to_string(), result);
+        }
+        report.overall_success = any_success;
```
adpaco: This means CI would pass even if 3/4 languages fail, right? What's the thought process here? The success condition should at least be configurable.

integration-tests/runner/src/aws.rs
```diff
@@ -0,0 +1,285 @@
...
+                debug!(
+                    "sts:AssumeRole attempt {}/{} failed for role '{}' ({}), retrying in {}s ...",
+                    attempt + 1,
+                    MAX_ATTEMPTS,
+                    target_role_name,
+                    e,
+                    wait
+                );
+                sleep(Duration::from_secs(wait)).await;
+                last_err = anyhow!("sts:AssumeRole failed for role '{target_role_name}': {e}");
```
adpaco: We shouldn't sleep or print the "retrying in..." message in the last attempt.

adpaco (commented):
Haven't finished my review, but here's a couple comments.

integration-tests/projects/run_001/cdk/bin/app.ts
mschlaipfer: Artifact from how these were initially generated. I renamed the runs to just have an ascending ID (and dropped the hash).

integration-tests/projects/run_003/cdk/bin/app.ts
mschlaipfer: This is an artifact from the original names they were generated as. Since these are throaway test stacks, the name here really doesn't matter. I didn't rename them when I renamed the run directory names, where these names would have been problematic.

integration-tests/projects/run_002/cdk/bin/app.ts
mschlaipfer: Same, I don't think these names matter.

integration-tests/runner/src/main.rs
mschlaipfer: Great catch, sorry for missing this.

mschlaipfer:
New run after latest fixes: https://github.com/mschlaipfer/iam-policy-autopilot/actions/runs/25791674475

janstey-amzn (dismissed):

integration-tests/runner/src/main.rs
```diff
@@ -0,0 +1,850 @@
...
+            }
+        }
+
+        // ── Step 3: CDK Destroy ─────────────────────────────────────────────
+        if cli.skip_destroy {
+            info!("[SKIP] CDK destroy (--skip-destroy)");
+            report.cdk_destroy = CdkStepResult::skipped();
+        } else {
+            let ok = cdk_destroy(&run_dir, deploy_role_arn, &cli.region, &sts).await?;
+            report.cdk_destroy = CdkStepResult::done(ok);
```
adpaco, line 423: I wonder if the overall success value should depend on CDK destroy.

Currently, if all languages pass then we report success, but any failures in the next steps could fail and be silently ignored or not report. In other words, I want to avoid a situation where the main tests are passing, but we're failing to clean-up and don't notice it.

integration-tests/runner/src/minimizer/mod.rs
```diff
@@ -0,0 +1,484 @@
...
+                        })
+                        .collect::<Vec<_>>()
+                })
+                .unwrap_or_default()
+        })
+        .collect();
+
+    let subset: Vec<Value> = indices.iter().map(|&i| all_policies[i].clone()).collect();
+    *runs_performed += 1;
+    warn!(
```
adpaco: Why don't we use `info!` on these messages instead?

.github/workflows/cdk-security-scan.yml
```diff
@@ -0,0 +1,73 @@
...
+jobs:
+  security-scan:
+    name: Checkov CDK Security Scan
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+      # ── 1. Check out repository ────────────────────────────────────────
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
adpaco: The other workflow uses `actions/checkout@v6`, should we use it here as well?

adpaco (dismissed):

adpaco (approved):

janstey-amzn (approved):

PR #203, opened 2026-05-15
test: fix silent JavaScript extraction test failures

adpaco:
*Description of changes:*

Converted the 5 JavaScript extraction integration tests into a single `#[rstest]` parameterized test with named cases. Removed the `Err` branch that silently accepted "Failed to load" errors as passing. If extraction fails, the test now correctly fails.


mschlaipfer (approved):
nice catch

janstey-amzn (approved):

PR #209, opened 2026-05-21
feat: use native certificate store

mschlaipfer:
*Issue #, if available:*

*Description of changes:*

We did not use the OS's native certificate store, but bundled certificates. This can cause issues with network traffic in monitored networks.

This PR also improves the error message to highlight the network request to the service endpoint URL failed.

For testability I added an env var for overriding the endpoint URL. I opted not to document it as part of this change, but we could in the future, in case users ask to host their own version of the service reference.

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
@@ -149,8 +149,8 @@ impl Engine {
                 }
                 Err(e) => {
                     return Err(ExtractorError::enrichment_error(
-                        &method.name,
-                        format!("Failed to enrich method call: {e}"),
+                        method.possible_services.join(", "),
+                        format!("Failed to enrich method call '{}': {e}", method.name),
                     ));
                 }
             }
--- a/iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
+++ b/iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
@@ -149,8 +149,9 @@ impl Engine {
                 }
                 Err(e) => {
                     return Err(ExtractorError::enrichment_error(
-                        method.possible_services.join(", "),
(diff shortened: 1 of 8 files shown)


janstey-amzn (dismissed):

iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
```diff
@@ -149,8 +149,8 @@ impl Engine {
                 }
                 Err(e) => {
                     return Err(ExtractorError::enrichment_error(
-                        &method.name,
-                        format!("Failed to enrich method call: {e}"),
+                        method.possible_services.join(", "),
+                        format!("Failed to enrich method call '{}': {e}", method.name),
```
adpaco: Nit: the constructor expects `service_name` (singular) and this would read a bit odd: ` Enrichment error for service 'iam, sts': Failed to enrich ...`.

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -426,12 +439,11 @@ impl RemoteServiceReferenceLoader {
...
-                ExtractorError::service_reference_parse_error_with_source(
-                    "RemoteServiceReferenceLoaderClientInitialization",
-                    "Failed to create service reference client".to_string(),
-                    e,
-                )
+            .map_err(|e| ExtractorError::Configuration {
+                message: format!(
+                    "Failed to initialize the HTTP client for the service reference endpoint: {e}"
+                ),
+                source: Some(Box::new(e)),
```
adpaco: We are formatting the error into the message so it will be duplicated.

adpaco (dismissed):

janstey-amzn (approved):

adpaco (approved):

PR #211, opened 2026-05-21
refactor: new extraction framework, implemented for Java

mschlaipfer:
*Issue #, if available:*

*Description of changes:*

Factor out a trait from the Java-specific extractor that serves as a framework for implementing future (or rewriting existing) language extractors. This should make adding support for new languages more structured, thus easier.

Changes the utility model slightly: now we have a two-tiered `HashMap` (service -> method -> utility), and optional `import` and `receiver` fields in the type (always `Some` for Java).


iam-policy-autopilot-policy-generation/src/extraction/framework/extractor_set.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /// Build a combined `any:` rule YAML from all registered extractors.
+    ///
+    /// Each extractor provides a rule body (content under `rule:`). These are assembled as
+    /// items under `rule:\n  any:`.
+    pub(crate) fn build_combined_rule(&self) -> String {
+        // Derive the display language name (PascalCase) from the type name for the rule id
+        // and language field. e.g. "ast_grep_language::Java" → "Java"
+        let lang_name = std::any::type_name::<L>()
```
adpaco, line 143: This is a bit fragile. Besides, I'm not sure it'd work with languages like C++ or C#.

iam-policy-autopilot-policy-generation/src/extraction/framework/language_extractor.rs
```diff
@@ -0,0 +1,136 @@
...
+///
+/// Fans out across `source_files` using `spawn_blocking` (CPU-bound tree-sitter work),
+/// merges per-file results via `IR::extend_from`, and returns the combined IR.
+pub(crate) async fn extract<E: LanguageExtractor>(
+    extractor: &E,
+    source_files: Vec<SourceFile>,
+) -> Result<E::ExtractionResult> {
+    extractor
+        .extractor_set()
+        .extract_from_files(source_files)
```
adpaco, line 134: We could initialize `extractor_set()` once and reuse in subsequent `extract` calls through a `LazyLock` or similar. Since our main workflows make a single extraction run, this is completely fine as is. But it'd be worth it if we ever need to run extraction on a loop.

adpaco (approved):

janstey-amzn (approved):

PR #217, opened 2026-05-26
feat: make resource cutoff configurable

### LINKED ISSUE #19: Make RESOURCE_CUTOFF configurable
We added functionality to prevent overly long resource lists added explicitly by resolving lists longer than 5 to `*` (https://github.com/awslabs/iam-policy-autopilot/pull/16). We should make the value configurable.

jmolz:
Issue #, if available:
Closes #19

Description of changes:

- Add `resource_cutoff` to the policy-generation config.
- Add `--resource-cutoff` to the CLI and optional `resource_cutoff` to the MCP input.
- Preserve the current default cutoff of 5 and add regression coverage for default/custom cutoff behavior.
- Update telemetry docs and the changelog for the new option.

Testing:

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo test -p iam-policy-autopilot-policy-generation resource_cutoff -- --nocapture`
- `cargo clippy --workspace -- -D warnings`
- `cargo test --workspace`


iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
```diff
@@ -67,7 +87,11 @@ impl Engine {
             .load_fas_maps_for_services(&unique_services, &service_cfg)
             .await?;
 
-        let resource_matcher = ResourceMatcher::new(service_cfg, fas_maps, sdk);
+        let resource_matcher = if self.resource_cutoff == crate::DEFAULT_RESOURCE_CUTOFF {
+            ResourceMatcher::new(service_cfg, fas_maps, sdk)
+        } else {
+            ResourceMatcher::with_resource_cutoff(service_cfg, fas_maps, sdk, self.resource_cutoff)
```
adpaco: Both branches do the same because you're calling `with_resource_cutoff` with the default value. We only need to call `with_resource_cutoff` here.

iam-policy-autopilot-mcp-server/src/tools/generate_policy.rs
```diff
@@ -101,6 +108,7 @@ pub async fn generate_application_policies(
         },
         aws_context: AwsContext::new(region, account)?,
         minimize_policy_size: false,
+        resource_cutoff: input.resource_cutoff.unwrap_or(DEFAULT_RESOURCE_CUTOFF),
```
adpaco: Could we validate this value just like the CLI does it? Otherwise the `resource_cutoff: 0` case will only be caught deeper in the stack.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Added
 
 - Support for namespace imports in TypeScript/JavaScript (#190)
+- Added a configurable resource cutoff for policy generation resource lists.
```
adpaco:
```suggestion
- Added a configurable resource cutoff for policy generation resource lists (#217)
```

TELEMETRY.md
```diff
@@ -22,6 +22,7 @@ Telemetry records which commands and parameters are used, and whether the comman
 | `upload_policies` | presence (boolean) |
 | `minimal_policy_size` | actual value (boolean) |
 | `disable_cache` | actual value (boolean) |
+| `resource_cutoff` | actual value (usize) |
```
adpaco: Doesn't this need the `omitted otherwise`?

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -363,6 +376,18 @@ Use this flag to force fresh data retrieval on every run."
...
         disable_cache: bool,
 
+        /// Resource-list cutoff before generated policies use wildcard resources
+        #[arg(
+            long = "resource-cutoff",
+            default_value_t = DEFAULT_RESOURCE_CUTOFF,
+            value_parser = parse_positive_usize,
+            long_help = "Number of enriched resources at which an action's resource list is \
+collapsed to '*' instead of emitting every resource-specific ARN. The default preserves the \
+previous behavior."
```
adpaco: Suggestion: Update the help text, the schema description and the `DEFAULT_RESOURCE_CUTOFF` with the following wording: `Resource lists with at least this many entries are collapsed to '*' instead of emitting every resource-specific ARN. Default: 5.`

The current documentation in those places is a bit ambiguous (i.e., does "at" mean "at exactly 5" or "starting at 5"?). There is also no need to refer to previous behavior.

iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
jmolz: Addressed in `ddaf5cb`: the enrichment engine now always calls `ResourceMatcher::with_resource_cutoff(...)`, so the redundant default branch is gone.

iam-policy-autopilot-mcp-server/src/tools/generate_policy.rs
jmolz: Addressed in `ddaf5cb`: the MCP tool now validates `resource_cutoff > 0` before building the generation config, and I added a regression test for `resource_cutoff: Some(0)`.

TELEMETRY.md
jmolz: Addressed in `ddaf5cb`: CLI telemetry now records `resource_cutoff` only when provided, and `TELEMETRY.md` says `value if provided, omitted otherwise`.

iam-policy-autopilot-cli/src/main.rs
jmolz: Addressed in `ddaf5cb`: the CLI help text, MCP schema description, and `DEFAULT_RESOURCE_CUTOFF` docs now use the suggested wording.

CHANGELOG.md
jmolz: Applied in `ddaf5cb`.

iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
```diff
@@ -31,8 +33,26 @@ impl Engine {
     /// Allows customization of the base paths for OperationAction maps and Service Reference files,
     /// useful for testing or alternative configurations.
     pub fn new(disable_file_system_cache: bool) -> Result<Self> {
+        Self::with_resource_cutoff(disable_file_system_cache, crate::DEFAULT_RESOURCE_CUTOFF)
+    }
+
+    /// Create a new MethodEnrichmentEngine with a custom resource cutoff.
+    ///
+    /// Resource lists with at least this many entries are collapsed to '*' instead of emitting every resource-specific ARN. Default: 5.
+    pub fn with_resource_cutoff(
```
mschlaipfer: Instead of introducing this, let's just change `new` to take another argument. I expect we'll need to update some call sites, but that seems better than introducing a new code path.

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -149,17 +149,43 @@ impl FasExpansion {
...
+    #[cfg(test)]
+    pub(crate) fn new(
+        service_cfg: Arc<ServiceConfiguration>,
+        fas_maps: OperationFasMaps,
+        sdk: SdkType,
+    ) -> Self {
+        Self::with_resource_cutoff(service_cfg, fas_maps, sdk, crate::DEFAULT_RESOURCE_CUTOFF)
+    }
+
+    pub(crate) fn with_resource_cutoff(
```
mschlaipfer: Let's just change `new`, same comment as on the `Engine`.

iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
jmolz: Addressed in 399500d: Engine::new now accepts resource_cutoff directly, and the separate with_resource_cutoff constructor is removed.

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
jmolz: Addressed in 399500d: ResourceMatcher::new now accepts resource_cutoff directly, and the separate with_resource_cutoff constructor is removed.

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -263,7 +280,7 @@ impl ResourceMatcher {
                                         &service_reference,
                                     )?;
                                 let enriched_resources =
-                                    if RESOURCE_CUTOFF <= enriched_resources.len() {
+                                    if self.resource_cutoff <= enriched_resources.len() {
```
mschlaipfer: I'm wondering why `resource_cutoff` must be `> 0`. Is it because otherwise we'd now add a `Resource: "*"`, even when an action doesn't operate on any resources (not sure if there are any such actions)?

jmolz: Yes, that's the reason. The cutoff is treated as an inclusive collapse threshold, so `0` would make `resource_cutoff <= enriched_resources.len()` true for every action, including an empty resource list. That would turn “no resource candidates” into `Resource: "*"`, and would also wildcard every single-resource action. I kept `0` invalid so the setting only represents a positive threshold rather than a special “always wildcard” mode.

iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
adpaco: Note that technically this is a breaking change. The enrichment engine is part of the public API so changing this constructor means we'll break anyone using it. That's probably fine as I doubt anyone is doing it at the moment, but it'd be nice to at least call out in the changelog via its own "Changed" line. Something like this:

```text
  ### Added
  - `--resource-cutoff` CLI flag and `resource_cutoff` MCP input to configure when resource lists collapse to `*` (#217)

  ### Changed
  - `EnrichmentEngine::new` now requires a `resource_cutoff` parameter; use `DEFAULT_RESOURCE_CUTOFF` to preserve existing behavior (#217)
```

jmolz: Addressed in 823ffe7: CHANGELOG.md now keeps the flag/MCP input under Added and calls out the EnrichmentEngine::new resource_cutoff parameter under Changed, with DEFAULT_RESOURCE_CUTOFF noted for existing behavior.

adpaco (dismissed):

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: We should use `<` in the comparison and allow `resource_cutoff := 0`. `> 0` is an invariant that we must enforce in multiple places, and I don't think it's needed.

#### Change to `<`

I believe a user setting `--resource-cutoff 1` would expect 1 resource to remain before falling back to `*`.

This means we'll need to change the default to `4` to maintain current behavior.

#### Allowing 0

`0` meaning "Always wildcard resources" seems perfectly fine, but maybe it's worth an extra call out in the help text, though.

When the resources are empty with `resource_cutoff == 0`: we pass on the empty vec, and fall back to `"Resource:": "*"` (a resource block is required) [here](https://github.com/awslabs/iam-policy-autopilot/blob/7a43447bf618314239b4d05f4e70ebb9ba2d43fb/iam-policy-autopilot-policy-generation/src/policy_generation/engine.rs#L148-L151)

#### Expected behavior

Can we add tests for the below?

| cutoff | resources | `cutoff < len` | result |
|--------|-----------|----------------|--------|
| 0 | 0 | false | empty vec → `"*"` (downstream fallback) |
| 4 | 4 | false | keeps all 4 |
| 4 | 5 | true | collapses to `"*"` |

jmolz: Addressed in 5fc1e3b: switched the cutoff comparison to `<`, changed the default to 4 to preserve existing behavior, removed the `> 0` validation so `0` is allowed, and updated the CLI/MCP/API wording to call out the `0` behavior. I also added regression coverage for the requested 0/0, 4/4, and 4/5 cases, plus a downstream empty-resource fallback test.

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 ### Added
 
 - Support for namespace imports in TypeScript/JavaScript (#190)
+- `--resource-cutoff` CLI flag and `resource_cutoff` MCP input to configure when resource lists collapse to `*` (#217)
+
+### Changed
+
+- `EnrichmentEngine::new` now requires a `resource_cutoff` parameter; use `DEFAULT_RESOURCE_CUTOFF` to preserve existing behavior (#217)
+- `resource_cutoff` now keeps exactly `resource_cutoff` resources, collapses larger non-empty resource lists to `*`, and accepts `0` to always wildcard non-empty resource lists (#217)
```
mschlaipfer: The entry in `Added` already covers this.

iam-policy-autopilot-cli/src/main.rs
```diff
@@ -363,6 +372,17 @@ Use this flag to force fresh data retrieval on every run."
         #[telemetry(value)]
         disable_cache: bool,
 
+        /// Resource lists with more than this many entries are collapsed to '*' instead of emitting every resource-specific ARN. Use 0 to collapse every non-empty resource list. Default: 4.
+        #[arg(
+            long = "resource-cutoff",
+            value_parser = parse_resource_cutoff,
```
mschlaipfer: Do we need this, if we don't enforce any custom constraints?

jmolz: Removed the custom parser and let Clap parse `usize` directly now that zero is accepted and no extra constraints remain.

CHANGELOG.md
jmolz: Removed the redundant `Changed` entries and left the resource-cutoff changelog note in `Added`.

iam-policy-autopilot-policy-generation/resources/config/sdks/boto3
mschlaipfer, line 1: The git submodule shouldn't change.

iam-policy-autopilot-policy-generation/resources/config/sdks/botocore-data
mschlaipfer, line 1: The submodule shouldn't change.

iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
mschlaipfer: Looks like my initial comment was never addressed?

iam-policy-autopilot-policy-generation/resources/config/sdks/boto3
jmolz: Resolved in `84fb759`: the `boto3` submodule gitlink is restored to the `upstream/main` SHA, so it no longer appears in the PR diff.

iam-policy-autopilot-policy-generation/resources/config/sdks/botocore-data
jmolz: Resolved in `84fb759`: the `botocore-data` submodule gitlink is restored to the `upstream/main` SHA, so it no longer appears in the PR diff.

iam-policy-autopilot-policy-generation/src/enrichment/engine.rs
jmolz: You were right — I missed this when preserving the old constructor path during the previous compatibility pass. Resolved in `84fb759`: `with_resource_cutoff` is removed, `EnrichmentEngine::new` now takes `resource_cutoff` directly, and the call sites/tests use `DEFAULT_RESOURCE_CUTOFF` where they need existing behavior.

iam-policy-autopilot-mcp-server/src/tools/generate_policy.rs
```diff
@@ -153,8 +165,9 @@ mod api {
     static MOCK_RETURN_VALUE: OnceLock<Mutex<Option<Result<GeneratePoliciesResult>>>> =
         OnceLock::new();
 
-    pub async fn generate_policies(
+    pub async fn generate_policies_with_options(
```
mschlaipfer: Let's keep the name `generate_policies`. That will also avoid the big diff in `tools/generate_policy.rs`

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -149,17 +149,29 @@ impl FasExpansion {
 /// This struct provides the core functionality for the 3-stage enrichment pipeline,
 /// combining parsed method calls with operation action maps and Service
 /// Definition Files to produce complete IAM metadata.
-#[derive(derive_new::new, Debug, Clone)]
```
mschlaipfer, line 152: Why can't we use `derive_new::new`?

mschlaipfer (commented):
thanks, getting close, just minor changes

iam-policy-autopilot-mcp-server/src/tools/generate_policy.rs
jmolz: Addressed in the current PR state: the MCP wrapper now keeps the local `generate_policies` name by aliasing `generate_policies_with_options` behind it, and the config construction is kept separate so the diff in `tools/generate_policy.rs` stays focused on adding the options pass-through.

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
jmolz: Good point. There was no custom constructor logic here, so I restored `#[derive(derive_new::new, Debug, Clone)]` and removed the manual `new` implementation. The generated constructor still takes `resource_cutoff` as the final field argument.

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -47,6 +47,21 @@ pub struct GeneratePolicyConfig {
     pub explain_resource_filters: Option<Vec<String>>,
 }
 
+/// Optional settings for policy generation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct GeneratePolicyOptions {
```
mschlaipfer: We don't need to introduce a new type for this, let's add the field to `GeneratePolicyConfig`?

jmolz: Adjusted in `52c4d37`: removed `GeneratePolicyOptions` and moved `resource_cutoff` onto `GeneratePolicyConfig`. The CLI and MCP now populate that config field directly and both call the existing `generate_policies` API.

adpaco (commented):
Thanks! You might want to remove this from the PR description

> validate it is greater than zero.

since 0 is now a valid value.

jmolz:
> Thanks! You might want to remove this from the PR description
> 
> > validate it is greater than zero.
> 
> since 0 is now a valid value.

Good catch, thanks. I updated the PR description to remove the stale “validate it is greater than zero” language now that 0 is a valid value.

mschlaipfer:
LGTM now, but it looks like not all your commits are signed.

jmolz:
Thanks for catching this. I rebuilt the branch to preserve the 13-commit PR history and re-signed each commit individually; GitHub now reports all 13 PR commits as verified.

jmolz:
Quick update: I rebased this branch onto current `main` to clear the dirty/conflicting state. New head is `2be039a`, and I verified all 13 PR commits are individually signed/verified. GitHub now reports the PR as mergeable again; the check rollup has not populated yet for the new head, so I will keep an eye on it.

jmolz:
Follow-up on the failed PR Checks run: the macOS arm64 failure was from the CLI integration tests reaching the live Service Reference endpoint:

`https://servicereference.us-east-1.amazonaws.com`

I pushed `5bae2a3` to seed minimal Service Reference cache fixtures in the CLI integration test harness before spawning `generate-policies`, so these tests no longer depend on live network reachability for Service Reference data.

Local validation:
- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo test -p iam-policy-autopilot-cli --test integration_tests -- --nocapture` (31 passed)
- `cargo clippy --workspace -- -D warnings`

I also verified GitHub reports all 14 PR commits as signed/verified. The new PR Checks run for `5bae2a3` is currently `action_required` with no jobs started: https://github.com/awslabs/iam-policy-autopilot/actions/runs/27133234388

CHANGELOG.md
```diff
@@ -11,12 +11,17 @@
...
+
+- `EnrichmentEngine::new` now requires a `resource_cutoff` parameter; use `DEFAULT_RESOURCE_CUTOFF` to preserve existing behavior (#217)
+
 ## [0.2.2rc1] - 2026-05-29
 
 ### Added
 
 - Support for namespace imports in TypeScript/JavaScript (#190)
 - Added partial support for permissions needed by [aws-lambda-powertools](https://pypi.org/project/aws-lambda-powertools/) (#186)
+- `--resource-cutoff` CLI flag and `resource_cutoff` MCP input to configure when resource lists collapse to `*` (#217)
```
adpaco: This line should be in the `Unreleased` section like the other one, no?

adpaco (commented):
What prompted you to include the last commit? If I'm not mistaken, the `generate_policy_command` function was already being used by other tests.

mschlaipfer:
My first instinct (haven't had time to look closely yet) also was that [5bae2a3](https://github.com/awslabs/iam-policy-autopilot/pull/217/commits/5bae2a313c6ab62f6525670643092687c84b0599) isn't needed, and was triggered by a transient failure to contact the service reference host (which shouldn't be addressed as part of this PR, if that's what it was).

CHANGELOG.md
jmolz: Agreed. I moved the `--resource-cutoff` / `resource_cutoff` entry into `Unreleased` and consolidated the duplicate `### Changed` heading in 9dcea1f.

jmolz:
@adpaco @mschlaipfer You were right about 5bae2a3. I added it after interpreting the previous macOS CI failure as a live Service Reference dependency in the CLI integration tests, but the rerun passing on the prior head 2be039a makes that look transient and out of scope for this PR. I force-pushed the branch to drop that commit and kept the follow-up scoped to the changelog placement fix in 9dcea1f. Local checks run: `git diff --check upstream/main...HEAD`, `cargo fmt --all -- --check`, and `cargo test -p iam-policy-autopilot-policy-generation resource_cutoff`.

adpaco (approved):

mschlaipfer (approved):

mschlaipfer:
Thanks for all the revisions

jmolz:
My pleasure. I enjoyed working through this with you.

PR #219, opened 2026-05-29
feat(wasm): Full Rust pipeline compiled to WebAssembly via Emscripten

janstey-amzn:
Closes # -

## Description

Compiles the entire IAM Policy Autopilot pipeline (AST extraction + enrichment + policy generation) to WebAssembly, producing a browser-ready npm package. Source code never leaves the browser.

This replaces the previous wasm-bindgen approach that split extraction (TypeScript) from generation (Rust WASM). The full pipeline now runs in a single WASM binary — single source of truth, no extraction parity concerns.

npm wrapper (generatePolicies) → loads Emscripten module + wraps with WebAssembly.promising() → calls generate_policies_wasm(json_input) → Rust: AST extraction (ast-grep + tree-sitter) → Rust: Enrichment (service reference via JSPI browser fetch) → Rust: Policy generation → returns JSON policy document

Async HTTP (service reference lookups) uses **JSPI** (JavaScript Promise Integration) — the WASM stack suspends during `fetch()` and resumes with the result. No Asyncify, no code size bloat.

## What's included

- **`iam-policy-autopilot-wasm/`** — new crate (staticlib → emcc → JS glue + .wasm)
- **`npm/`** — publishable npm package with `generatePolicies()` API and `checkBrowserSupport()`
- **`build.sh`** — three-stage build: cargo → emcc → tsc
- **`.github/workflows/wasm.yml`** — CI workflow for the WASM build
- **`HttpGet` trait** — pluggable HTTP backend (reqwest on native, browser fetch via FFI on wasm)
- **Platform gating via `cfg(target_arch = "wasm32")`** — automatic, no feature flags needed

## Browser requirements

- Chrome 131+ or Chromium-based (JSPI required)
- CSP: `unsafe-eval` + `connect-src` for servicereference.amazonaws.com
- Firefox/Safari not supported (no JSPI)

## Binary size

~55MB uncompressed, **~6.5MB gzipped**. Includes all tree-sitter grammars (~25 languages) but only 5 are used. `ast-grep-language` doesn't support selective grammar inclusion without a fork — tracked as follow-up work.

## Verified

- Native build: zero warnings, all tests pass ✓
- Emscripten build: zero warnings ✓
- E2E in Vite+React test app ✓
- clippy, fmt, cargo-machete all clean ✓

## Checklist

- [ ] I have updated [`CHANGELOG.md`](https://github.com/awslabs/iam-policy-autopilot/blob/main/CHANGELOG.md) under `[Unreleased]` (if this is a user-facing change)

---


Cargo.toml
```diff
@@ -86,7 +87,9 @@ strsim = "0.11"
 
 # Additional workspace configuration
 [workspace.lints.rust]
-unsafe_code = "forbid"
+# deny (not forbid) so the emscripten WASM crate can #[allow(unsafe_code)] for FFI.
+# All other crates should avoid unsafe code.
+unsafe_code = "deny"
```
adpaco, line 94: We're already overriding in the `iam-policy-autopilot-wasm` crate, why do we change to `deny` here? We should keep the lint for other crates.

iam-policy-autopilot-wasm/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Accepts a JSON string, runs extraction + enrichment + policy generation,
+/// returns a JSON string with the result. Caller must free the returned
+/// pointer with `free_string`.
+///
+/// # Safety
+/// - `input_ptr` must be a valid, non-null, null-terminated C string.
+/// - The caller must free the returned pointer with `free_string`.
+#[no_mangle]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
+pub extern "C" fn generate_policies_wasm(input_ptr: *const c_char) -> *mut c_char {
```
adpaco: This function should be declared `unsafe`.

iam-policy-autopilot-wasm/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+}
+
+/// Free a string previously returned by `generate_policies_wasm`.
+///
+/// # Safety
+/// `ptr` must have been returned by `generate_policies_wasm` and must not be freed twice.
+#[no_mangle]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
+pub extern "C" fn free_string(ptr: *mut c_char) {
```
adpaco: This function should be declared `unsafe`.

iam-policy-autopilot-wasm/npm/src/index.ts
```diff
@@ -0,0 +1,212 @@
...
+  locateFile?: (filename: string) => string;
+}
+
+/**
+ * Initialize the WASM module. Called automatically on first use,
+ * but can be called explicitly to pre-load the module.
+ */
+export async function init(options?: InitOptions): Promise<void> {
+  if (modulePromise) return;
+  modulePromise = loadModule(options);
```
adpaco, line 123: If `loadModule` fails for whatever reason, we will never attempt to reload it.

iam-policy-autopilot-wasm/npm/src/index.ts
```diff
@@ -0,0 +1,212 @@
...
+ * AWS Service Reference endpoint for IAM action metadata.
+ *
+ * @param input - Source files and optional AWS context
+ * @returns Generated IAM policies
+ * @throws Error if the WASM module fails to load or generation encounters an unrecoverable error
+ */
+export async function generatePolicies(input: GenerateInput): Promise<GenerateResult> {
+  if (!modulePromise) {
+    modulePromise = loadModule();
+  }
```
adpaco: Same issue as above. This should probably just delegate to `init()`?

iam-policy-autopilot-policy-generation/src/enrichment/http_client.rs
```diff
@@ -0,0 +1,172 @@
...
+
+            if result_ptr.is_null() {
+                return Err(ExtractorError::Network {
+                    message: format!("Fetch failed for '{url}'"),
+                    source: None,
+                });
+            }
+
+            // Read the C string, then free immediately to avoid leaks on error paths.
+            let c_str = unsafe { std::ffi::CStr::from_ptr(result_ptr as *const i8) };
```
adpaco: It's not an issue for this platform, but generally `c_char` is not `i8` on all platforms.
```suggestion
            let c_str = unsafe { std::ffi::CStr::from_ptr(result_ptr as *const c_char) };
```

iam-policy-autopilot-wasm/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+    {
+        Ok(rt) => rt,
+        Err(e) => {
+            return serde_json::json!({"error": format!("Failed to create async runtime: {e}")})
+                .to_string();
+        }
+    };
+
+    let result = rt.block_on(async {
+        let aws_context = AwsContext::new(input.region, input.account)?;
```
adpaco, line 163: What about partition?

iam-policy-autopilot-policy-generation/src/embedded_data.rs
```diff
@@ -138,23 +138,8 @@ impl JavaSdkUtilitiesRaw {
 impl Boto3ResourcesRaw {
     /// Get a boto3 resources definition file by service name and API version
     fn get_resources_definition(service: &str, api_version: &str) -> Option<Cow<'static, [u8]>> {
-        let start_time = std::time::Instant::now();
```
adpaco, line 141: Why is this code being removed? Could we just config it out with `#[cfg(not(target_arch = "wasm32"))]`?

iam-policy-autopilot-policy-generation/src/api/generate_policies_from_source.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    let result = policy_engine
+        .generate_policies(&enriched)
+        .context("Failed to generate IAM policies")?;
+
+    let merged = policy_engine
+        .merge_policies(&result.policies)
+        .context("Failed to merge IAM policies")?;
+
+    Ok(GeneratePoliciesResult::new(merged, None))
```
adpaco: Is there any reason not to include explanations? Need at least a comment here.

iam-policy-autopilot-wasm/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+    #[serde(default = "default_wildcard")]
+    region: String,
+    /// AWS account ID for ARN generation. Defaults to "*".
+    #[serde(default = "default_wildcard")]
+    account: String,
+    /// Optional language override (auto-detected from filename if omitted).
+    language: Option<String>,
+    /// Enable minimal policy size by allowing cross-service merging. Defaults to false.
+    #[serde(default)]
+    minimize_policy_size: bool,
```
adpaco, line 47: Do we want `minimizePolicySize` (camelCase)?

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          submodules: true
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-emscripten
+
+      - uses: mymindstorm/setup-emsdk@v14
+        with:
+          version: latest
```
adpaco: Not sure we should use the latest.

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-emscripten
+
+      - uses: mymindstorm/setup-emsdk@v14
```
mschlaipfer: What is `mymindstorm`? Can't find that.

iam-policy-autopilot-policy-generation/resources/config/sdks/boto3
mschlaipfer, line 1: submodules shouldn't change as part of this PR

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
```diff
@@ -343,16 +335,16 @@ pub async fn generate_policies(config: &GeneratePolicyConfig) -> Result<Generate
...
         final_policies.len(),
     );
 
-    Ok(GeneratePoliciesResult {
-        policies: final_policies,
-        explanations,
-        resource_binding_explanations: binding_explanations,
-    })
+    let mut result = GeneratePoliciesResult::new(final_policies, explanations);
+    result.resource_binding_explanations = binding_explanations;
```
mschlaipfer: This doesn't seem better. Adding `new` seems fine though, if we don't have it yet.

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -61,6 +61,18 @@ pub struct GeneratePoliciesResult {
     pub resource_binding_explanations: Option<Vec<ResourceBindingExplanation>>,
 }
 
+impl GeneratePoliciesResult {
+    /// Create a new result with policies and optional explanations.
+    #[must_use]
+    pub fn new(policies: Vec<PolicyWithMetadata>, explanations: Option<Explanations>) -> Self {
```
mschlaipfer: Seems arbitrary to exclude `resource_binding_explanations`. How about we derive a builder instead?

iam-policy-autopilot-policy-generation/src/api/common.rs
```diff
@@ -48,6 +48,7 @@ pub(crate) async fn process_source_files(
     let language = Language::try_from_str(&language)?;
 
     // Record detected language into telemetry span (no-ops if no scope active)
+    #[cfg(feature = "telemetry")]
```
mschlaipfer, line 51: The PR description doesn't mention anything about telemetry. What's the plan for Wasm?

iam-policy-autopilot-policy-generation/src/api/generate_policies_from_source.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Enable minimal policy size by allowing cross-service merging.
+    pub minimize_policy_size: bool,
+}
+
+/// Generate IAM policies from in-memory source code.
+///
+/// This is the WASM-friendly counterpart of [`super::generate_policies`].
+/// It skips filesystem I/O entirely — source content is provided directly
+/// and the Service Reference filesystem cache is disabled.
+pub async fn generate_policies_from_source(
```
mschlaipfer, line 35: Can we avoid mostly duplicating this logic from the existing API?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
mschlaipfer, line 1: This file has 23 mentions of `wasm32`. Not sure what the best way is, but would using modules work to make this maintainable?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -307,13 +323,80 @@ fn deserialize_service_reference_mapping(
 #[derive(Debug)]
 
 pub(crate) struct RemoteServiceReferenceLoader {
-    client: Client,
```
mschlaipfer, line 310: Why is this dynamic dispatch at runtime and not conditionally compiled?

iam-policy-autopilot-policy-generation/Cargo.toml
```diff
@@ -14,31 +14,41 @@ crate-type = ["cdylib", "rlib"]
...
+
+# Native-only dependencies (not available on WASM)
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 async-lsp = { version = "0.2", features = ["tokio"] }
 lsp-types = "0.95"
 tower = "0.5"
 which = "6.0"
+tokio-util = { workspace = true }
+rust-embed = { workspace = true }
+reqwest = { workspace = true }
```
mschlaipfer: NP: Let's use the `.workspace = true` style for consistency

iam-policy-autopilot-wasm/build.sh
mschlaipfer, line 1: Can this be a `build.rs` or can this follow the xtask pattern already used in the project so we can write script like this in typed Rust?

iam-policy-autopilot-policy-generation/Cargo.toml
```diff
@@ -14,31 +14,41 @@ crate-type = ["cdylib", "rlib"]
...
 tokio = { workspace = true, features = ["process", "io-util"] }
-tokio-util.workspace = true
-hcl-rs.workspace = true
-walkdir.workspace = true
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+rust-embed = { version = "8.11", features = ["include-exclude"] }
+tokio = { version = "1.0", features = ["rt", "macros", "sync"] }
```
mschlaipfer, line 52: Can this use the workspace version?

iam-policy-autopilot-wasm/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+crate-type = ["staticlib"]
+
+[dependencies]
+iam-policy-autopilot-policy-generation = { path = "../iam-policy-autopilot-policy-generation", default-features = false }
+
+serde = { workspace = true }
+serde_json = { workspace = true }
+# Workspace tokio includes fs/rt-multi-thread/signal which pull in mio's net features,
+# incompatible with wasm32. We only need the single-threaded runtime here.
+tokio = { version = "1.0", features = ["rt", "macros"] }
```
mschlaipfer, line 19: Can this use the workspace version?

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          version: latest
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+
+      - name: Build WASM + npm package
+        run: ./iam-policy-autopilot-wasm/build.sh
```
mschlaipfer, line 39: This should just be a `cargo build` command when we move the `build.sh` to `build.rs`.

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Verify artifacts exist
+        run: |
+          test -f iam-policy-autopilot-wasm/npm/dist/iam_policy_autopilot.js
+          test -f iam-policy-autopilot-wasm/npm/dist/iam_policy_autopilot.wasm
+          test -f iam-policy-autopilot-wasm/npm/dist/index.js
+          test -f iam-policy-autopilot-wasm/npm/dist/index.d.ts
+          echo "All artifacts present"
+          ls -lh iam-policy-autopilot-wasm/npm/dist/
+
+      - name: Upload npm package artifact
```
mschlaipfer, line 53: This releases to NPM? What will the landing page look like? Does it use one of our READMEs?

iam-policy-autopilot-policy-generation/src/providers/mod.rs
mschlaipfer, line 1: It would be awesome if we could use conditionally compiled "providers" (maybe runtimes is a better name), so we can keep the same core logic, and all the wasm/native distinction lives in this crate. We can use a linter to ensure the wrapped APIs aren't used anywhere besides this module.

It might be worth trying to see if an agent can get this done, as it would be a vastly improved design.

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,55 @@
...
+  build-wasm:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-emscripten
```
mschlaipfer, line 28: What are the tradeoffs between the different Wasm runtimes?

iam-policy-autopilot-wasm/src/lib.rs
janstey-amzn: The npm package targets public internet users who shouldn't need to know about AWS partitions — it's an internal concept. If we ever vend this internally, we can add it as an optional field.

.github/workflows/wasm.yml
janstey-amzn: It's the standard GitHub Action for installing emsdk: https://github.com/mymindstorm/setup-emsdk. Used by emscripten-core's own CI and widely adopted.

.github/workflows/wasm.yml
janstey-amzn: A build.rs runs before compilation, but the emcc link step needs the compiled .a as input (post-compilation). Plan to migrate to cargo xtask wasm-build (already have an xtask crate in the workspace) as a follow-up, which gives us Rust-native build orchestration without the ordering constraint.

.github/workflows/wasm.yml
janstey-amzn: No, upload-artifact just saves the build output for download from the workflow run. There's no npm publish step, that will be a separate release workflow with proper versioning and auth.

iam-policy-autopilot-policy-generation/resources/config/sdks/boto3
janstey-amzn: Fixed. Submodule pointers are now aligned with upstream/main. The diff previously showed them as changed because the branch was behind upstream when the PR was opened. After merging upstream/main into the branch, the submodules match.

iam-policy-autopilot-policy-generation/src/api/common.rs
janstey-amzn: Telemetry is disabled at compile time for WASM (default-features = false excludes the telemetry feature). The telemetry crate relies on filesystem-based persistent storage for configuration (opt-in state, client ID), which isn't available in the browser. If we want browser telemetry in the future, it would need a separate storage backend (e.g., localStorage). Tracked as follow-up.

.github/workflows/wasm.yml
janstey-amzn: Emscripten provides a full libc, filesystem emulation, pthreads shim, and a C compiler (emcc). tree-sitter (our AST parser) is C. Without it, we'd have to split extraction into JS (like the original approach) and only run enrichment+generation in WASM. Emscripten lets us compile the entire pipeline as one binary with a single source of truth, no extraction parity bugs.

iam-policy-autopilot-policy-generation/src/enrichment/http_client.rs
janstey-amzn: Done

iam-policy-autopilot-policy-generation/src/providers/mod.rs
janstey-amzn: Done

iam-policy-autopilot-policy-generation/Cargo.toml
janstey-amzn: Done

iam-policy-autopilot-policy-generation/Cargo.toml
janstey-amzn: No, the workspace tokio includes `fs`, `rt-multi-thread`, and signal features which depend on mio's network I/O, incompatible with `wasm32-unknown-emscripten`. The WASM target only needs the single-threaded runtime (`rt`, `macros`, `sync`). Added a comment above the line explaining this.

iam-policy-autopilot-wasm/npm/src/index.ts
janstey-amzn: Fixed

iam-policy-autopilot-wasm/build.sh
janstey-amzn: A build.rs runs before compilation, but the emcc link step needs the compiled .a as input (post-compilation). Plan to migrate to `cargo xtask wasm-build` (already have an `xtask` crate in the workspace) as a follow-up, which gives us Rust-native build orchestration without the ordering constraint.

iam-policy-autopilot-wasm/Cargo.toml
janstey-amzn: Same constraint as the policy-generation crate — workspace tokio includes `fs`, `rt-multi-thread`, and signal which depend on mio's net I/O, incompatible with `wasm32`. Comment above the line explains this.

Cargo.toml
janstey-amzn: `forbid` can't be used here. The policy-generation crate (not just the wasm crate) also contains `#[allow(unsafe_code)]` on a cfg-gated module (http_client.rs. The Emscripten FFI bridge). When Cargo passes forbid as `-F unsafe_code` on the command line, no `#[allow]` attribute anywhere in that crate can override it. Since policy-generation uses `[lints] workspace = true` (inheriting the workspace lint), forbid causes a hard compilation error on the `wasm32` target. Deny is the strongest level that allows scoped `#[allow]` for the FFI boundary.

iam-policy-autopilot-policy-generation/src/api/generate_policies_from_source.rs
janstey-amzn: Yeah, its a little bit of a larger change. I will do this in a followup PR

.github/workflows/wasm.yml
mschlaipfer: I think it should be
```
emscripten-core/setup-emsdk@v16
```
copied from https://github.com/marketplace/actions/setup-emscripten-toolchain.

Maybe `mymindstorm` is a fork.

.github/workflows/wasm.yml
mschlaipfer: What are the downsides? Binary size?

iam-policy-autopilot-policy-generation/src/providers/concurrency.rs
```diff
@@ -0,0 +1,58 @@
...
+    for task in tasks {
+        join_set.spawn(async move { task().await });
+    }
+
+    let mut results = Vec::new();
+    while let Some(result) = join_set.join_next().await {
+        match result {
+            Ok(value) => results.push(value),
+            Err(e) => {
+                log::error!("Task panicked: {e}");
```
mschlaipfer: NP: a panic wouldn't be caught here, I think?

iam-policy-autopilot-policy-generation/src/extraction/sdk_model.rs
```diff
@@ -303,106 +305,139 @@ impl ServiceDiscovery {
...
+            }
         }
 
-        log::debug!(
-            "Spawned all {} service loading tasks in {:?}",
-            join_set.len(),
-            start_time.elapsed()
-        );
+        // On native: load in parallel with JoinSet
+        #[cfg(not(target_arch = "wasm32"))]
```
mschlaipfer: Should this use `run_all` or maybe a more general version that can support this use case?

iam-policy-autopilot-policy-generation/src/providers/mod.rs
mschlaipfer: I don't think this is "Done". The providers aren't used throughout the code base, unfortunately.

For this to be "Done" we'd need to replace existing calls to native APIs with the provider APIs which abstract them (and use linter rules that forbid calling the native APIs directly).

iam-policy-autopilot-policy-generation/src/lib.rs
```diff
@@ -31,13 +30,15 @@ pub mod policy_generation;
...
+#[cfg(not(target_arch = "wasm32"))]
 pub mod lsp;
 
 use std::fmt::Display;
 use std::path::PathBuf;
 
 pub use enrichment::{Engine as EnrichmentEngine, Explanation};
-pub use extraction::{Engine as ExtractionEngine, ExtractedMethods, SdkMethodCall, SourceFile};
+pub use extraction::{Engine as ExtractionEngine, ExtractedMethods};
+pub use extraction::{SdkMethodCall, SourceFile};
```
mschlaipfer: NP: no need for this change

iam-policy-autopilot-policy-generation/src/lib.rs
```diff
@@ -48,6 +49,7 @@ pub use policy_generation::{
 // Re-export commonly used types for convenience
 pub(crate) use extraction::ServiceModelIndex;
 
+#[cfg(not(target_arch = "wasm32"))]
```
mschlaipfer: These providers should be implemented such that they can be used regardless of compile target.

iam-policy-autopilot-wasm/npm/src/index.ts
```diff
@@ -0,0 +1,216 @@
...
+export interface GenerateResult {
+  Policies: Policy[];
+}
+
+export interface GenerateError {
+  error: string;
+}
+
+type EmscriptenModule = {
+  _generate_policies_wasm: (ptr: number) => number;
```
mschlaipfer, line 63: What do names starting with underscore mean?

iam-policy-autopilot-wasm/npm/src/index.ts
```diff
@@ -0,0 +1,216 @@
...
+ * ```
+ */
+export function checkBrowserSupport(): { supported: boolean; missing: string[] } {
+  const missing: string[] = [];
+
+  if (typeof WebAssembly === 'undefined') {
+    missing.push('WebAssembly');
+  } else {
+    if (typeof (WebAssembly as any).promising !== 'function') {
+      missing.push('WebAssembly.promising (JSPI) — requires Chrome 131+ or a Chromium-based browser');
```
mschlaipfer: Why are we talking only about Chromium here? What about other browsers?

iam-policy-autopilot-wasm/npm/src/index.ts
```diff
@@ -0,0 +1,216 @@
...
+    throw e;
+  }
+}
+
+async function loadModule(options?: InitOptions): Promise<EmscriptenModule> {
+  // Check JSPI support before attempting to load
+  const support = checkBrowserSupport();
+  if (!support.supported) {
+    throw new Error(
+      `Browser does not support required WebAssembly features: ${support.missing.join('; ')}. ` +
```
mschlaipfer: Doesn't the missing message already contain an error string talking about the browser?

iam-policy-autopilot-wasm/npm/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "iam-policy-autopilot-wasm",
+  "version": "0.2.1",
```
mschlaipfer: How are we versioning this? Do we need to update our `RELEASE.md` runbook? Should this be `0.2.3`?

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -63,6 +63,59 @@ pub struct GeneratePoliciesResult {
...
+            explanations: None,
+            resource_binding_explanations: None,
+        }
+    }
+}
+
+/// Builder for [`GeneratePoliciesResult`].
+///
+/// Created via [`GeneratePoliciesResult::builder()`].
+pub struct GeneratePoliciesResultBuilder {
```
mschlaipfer: Let's use https://crates.io/crates/derive_builder

.github/workflows/wasm.yml
janstey-amzn: A few extra KB of overhead, ~20kb - 30kb, unzipped

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -3,25 +3,27 @@
...
+use std::collections::HashMap;
+#[cfg(not(target_arch = "wasm32"))]
+use std::path::PathBuf;
+use std::time::{Duration, SystemTime};
+#[cfg(not(target_arch = "wasm32"))]
 use tokio::fs;
 use tokio::sync::{OnceCell, RwLock};
 
 type OperationName = String;
+#[cfg(not(target_arch = "wasm32"))]
```
mschlaipfer: Why is this unused in Wasm?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -269,12 +271,12 @@ where
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) struct ServiceReferenceMapping {
     // represents the top level service reference mapping
-    pub(crate) service_reference_mapping: HashMap<String, Url>,
+    pub(crate) service_reference_mapping: HashMap<String, String>,
```
mschlaipfer, line 272: Why is the type changing?

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -50,19 +51,33 @@ pub struct GeneratePolicyConfig {
...
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Clone, Serialize, Builder)]
 #[serde(rename_all = "PascalCase")]
+#[builder(setter(into, strip_option))]
 pub struct GeneratePoliciesResult {
     /// Generated IAM policies
     pub policies: Vec<PolicyWithMetadata>,
     /// Explanations for why actions were added (if requested)
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[builder(default, setter(strip_option))]
```
mschlaipfer: Why strip_option? Makes it awkward to call?

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
janstey-amzn: The url crate was removed when we extracted the HttpGet trait (reqwest brought it transitively). We only ever use these URLs as base strings for string concatenation, so full URL parsing is unnecessary. The deserializer still validates the scheme.

iam-policy-autopilot-policy-generation/src/api/model.rs
janstey-amzn: `strip_option` means callers write `.explanations(explanations)` rather than `.explanations(Some(explanations))`. The field defaults to None if not set, so the Option wrapping is handled internally. It makes the builder API cleaner at call sites.

iam-policy-autopilot-wasm/npm/src/index.ts
janstey-amzn: Emscripten's C name-mangling convention: exported C symbols get a leading underscore. So `generate_policies_wasm` in Rust becomes `_generate_policies_wasm` in the JS glue. This is standard Emscripten behavior.

iam-policy-autopilot-wasm/npm/src/index.ts
janstey-amzn: You're right, the missing entries already include browser version details. I'll remove the redundant suffix.

iam-policy-autopilot-policy-generation/src/lib.rs
janstey-amzn: FileSystemProvider uses `tokio::fs` for reading files from disk which is not available on WASM. WASM consumers pass source code directly via the JSON API. The concurrency and json providers are already target-agnostic. I'll add a doc comment clarifying that FileSystemProvider is intentionally native-only

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
janstey-amzn: It's only used by the filesystem cache methods which are native-only. Moving it inside the cfg block to make the scoping clearer.

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -50,19 +51,33 @@ pub struct GeneratePolicyConfig {
 }
 
 /// Result of policy generation including policies, action mappings, and explanations
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Clone, Serialize, Builder)]
 #[serde(rename_all = "PascalCase")]
+#[builder(setter(into, strip_option))]
```
mschlaipfer: We already use `strip_option` on the fields, why on the `struct` as well?

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
```diff
@@ -344,16 +336,24 @@ pub async fn generate_policies(config: &GeneratePolicyConfig) -> Result<Generate
...
-        resource_binding_explanations: binding_explanations,
-    })
+    let mut builder = GeneratePoliciesResultBuilder::default();
+    builder.policies(final_policies.clone());
+    if let Some(explanations) = explanations {
+        builder.explanations(explanations);
+    }
+    if let Some(binding_explanations) = binding_explanations {
+        builder.resource_binding_explanations(binding_explanations);
+    }
```
mschlaipfer, line 394: Here `strip_option` makes using the builder awkward. Not sure if the explanations should be `Option`-typed

iam-policy-autopilot-policy-generation/src/api/model.rs
mschlaipfer: Okay, seems fine. I just want to highlight that it makes it both worse in some cases and better in others. Commented where it's worse. Not sure why we return the explanations as an `Option` (which makes the builder worse), maybe that could be changed, but not for this PR.

iam-policy-autopilot-wasm/src/lib.rs
mschlaipfer: I think we don't take the partition anymore, we infer it from the mapping provided by botocore.

.github/workflows/wasm.yml
```diff
@@ -31,7 +31,8 @@ jobs:
         with:
           version: 5.0.7
 
-      # Node 24+ needed for JSPI (--experimental-wasm-jspi) in the boundary test
+      # Node 24 needs --experimental-wasm-jspi for the boundary test.
+      # Node 26+ enables JSPI by default; drop the flag when bumping node-version.
```
mschlaipfer: Any reason not to bump this?

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
```diff
@@ -143,12 +141,7 @@ fn filter_resource_explanations(
 }
 
 /// Filter explanations to only include actions matching the given patterns.
-fn filter_explanations(
-    explanations: Option<Explanations>,
-    filters: &[String],
-) -> Option<Explanations> {
-    let explanations = explanations?;
-
+fn filter_explanations(explanations: Explanations, filters: &[String]) -> Explanations {
```
mschlaipfer: Can we now implement this in `impl Explanations`?

iam-policy-autopilot-wasm/npm/package.json
```diff
@@ -1,7 +1,11 @@
 {
-  "name": "iam-policy-autopilot-wasm",
+  "name": "@aws/iam-policy-autopilot",
   "version": "0.2.3",
-  "description": "Browser-ready WebAssembly build of IAM Policy Autopilot — generates least-privilege IAM policies from source code. Requires JSPI support (Chrome 137+, Firefox 153+). Not yet optimized for production use (large binary).",
+  "description": "Browser-ready WebAssembly build of IAM Policy Autopilot — generates IAM policies scoped to the AWS SDK calls in your source code. Requires JSPI support (Chrome/Edge 137+, Firefox 153+, Safari 27+). Not yet optimized for production use (large binary).",
```
mschlaipfer: How are we planning to address this?
> Not yet optimized for production use (large binary)

What's the binary size?

mschlaipfer (dismissed):

iam-policy-autopilot-wasm/npm/package.json
janstey-amzn: ~50MB

mschlaipfer (approved):

PR #229, opened 2026-06-08
feat(policy-generation): sort policy statements by service for deterministic output

### LINKED ISSUE #153: Improve policy output: merge statements with equivalent resources and sort statements by service name
Take the following policy

```
{
  "Policies": [
    {
      "Policy": {
        "Id": "IamPolicyAutopilot",
        "Version": "2012-10-17",
        "Statement": [
          {
            "Effect": "Allow",
            "Action": [
              "kms:Decrypt",
              "kms:GenerateDataKey"
            ],
            "Resource": [
              "arn:aws:kms:*:*:key/*"
            ],
            "Condition": {
              "StringLike": {
                "kms:ViaService": [
                  "s3.*.amazonaws.com"
                ]
              }
            }
          },
          {
            "Effect": "Allow",
            "Action": [
              "s3:GetObject",
              "s3:GetObjectAcl",
              "s3:GetObjectLegalHold",
              "s3:GetObjectRetention",
              "s3:GetObjectTagging",
              "s3:GetObjectVersion",
              "s3:PutObject",
              "s3:PutObjectAcl",
              "s3:PutObjectLegalHold",
              "s3:PutObjectRetention",
              "s3:PutObjectTagging"
            ],
            "Resource": [
              "arn:aws:s3:*:*:accesspoint/*/object/*",
              "arn:aws:s3:::*/*"
            ]
          },
          {
            "Effect": "Allow",
            "Action": [
              "s3-object-lambda:GetObject",
              "s3-object-lambda:PutObject"
            ],
            "Resource": [
              "arn:aws:s3:*:*:accesspoint/*/object/*",
              "arn:aws:s3:::*/*"
            ]
          },
          {
            "Effect": "Allow",
            "Action": [
              "dynamodb:PutItem",
              "dynamodb:Query",
              "dynamodb:UpdateItem",
              "dynamodb:WriteDataForReplication"
            ],
            "Resource": [
              "arn:aws:dynamodb:*:*:table/*"
            ]
          },
          {
            "Effect": "Allow",
            "Action": [
              "kms:Decrypt"
            ],
            "Resource": [
              "arn:aws:kms:*:*:key/*"
            ],
            "Condition": {
              "StringLike": {
                "kms:ViaService": [
                  "dynamodb.*.amazonaws.com"
                ]
              }
            }
          },
          {
            "Effect": "Allow",
            "Action": [
              "s3:ListBucket"
            ],
            "Resource": [
              "arn:aws:s3:*:*:accesspoint/*",
              "arn:aws:s3:::*"
            ]
          },
          {
            "Effect": "Allow",
            "Action": [
              "s3-object-lambda:ListBucket"
            ],
            "Resource": [
              "arn:aws:s3:*:*:accesspoint/*",
              "arn:aws:s3:::*"
            ]
          }
        ]
      },
      "PolicyType": "Identity"
    }
  ]
}
```

We should output all `kms` and all `s3` statements next to one another, respectively to make the policies easier to read/review.
Moreover, the `s3-object-lambda` statements could be merged with the `s3` statements since they operate on the same resources, without impact on the permissions granted by the policy.

valter-silva-au:
## Problem

Generated IAM policies emit statements in insertion order, so statements for the same service end up scattered throughout the policy. This hurts human review and produces unstable diffs across runs.

Issue #153 asks for two things: (1) merge statements with equivalent resources — already implemented — and (2) sort statements by service name — absent until now.

## Approach

Added a panic-free helper `sort_statements_by_service(&mut [Statement])` and call it on each policy's statements just before they are returned from `merge_statements`. Because both the single-policy path and the multi-policy-split path flow through `merge_statements`, this single call site covers both. The sort uses a stable composite key:

1. Service prefix of the first action (e.g. `s3` from `s3:GetObject`)
2. The full action list (tie-breaker between same-service statements)
3. The resource list (final tie-breaker)

An empty action list is treated as an empty-string service key, so the sort never panics on a malformed statement (no `unwrap`/`expect`). Within-statement action/resource sorting was already in place and is unchanged.

## Scope note

This PR is intentionally scoped to the deterministic sort (#153 item 2) only. Cross-service consolidation (e.g. merging `s3` and `s3-object-lambda` statements) touches permission semantics and needs a maintainer design decision, so it is left as a deliberate follow-up. The "merge equivalent resources" half of #153 already exists.

## Testing

- New unit test `test_statements_sorted_by_service` feeds scrambled services (s3, dynamodb, kms) and asserts the output is ordered by service prefix, same-service statements are adjacent, and the order is stable across repeated runs.
- `cargo test --workspace` — all green (32 suites, 0 failed).
- `cargo clippy --workspace -- -D warnings` — clean (matches the repo's CI gate).
- `cargo fmt --all --check` — clean.

Fixes #153

---

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/policy_generation/merge.rs
+++ b/iam-policy-autopilot-policy-generation/src/policy_generation/merge.rs
@@ -243,20 +243,23 @@ impl PolicyMerger {
     /// action's service. The "same service stays together" property therefore
     /// holds only when cross-service merging is off (the default).
     fn sort_statements_by_service(statements: &mut [Statement]) {
-        fn service(statement: &Statement) -> &str {
-            statement
-                .action
-                .first()
-                .and_then(|action| action.split(':').next())
-                .unwrap_or("")
-        }
-
         statements.sort_by(|a, b| {
-            service(a)
-                .cmp(service(b))
+            Self::service_prefix(a)
+                .cmp(Self::service_prefix(b))
                 .then_with(|| a.action.cmp(&b.action))
                 .then_with(|| a.resource.cmp(&b.resource))
         });
+    }
+
+    /// Service prefix of a statement's first action (e.g. `s3` from
+    /// `s3:GetObject`). Returns an empty string when the statement has no
+    /// actions, so callers never panic on malformed statements.
+    fn service_prefix(statement: &Statement) -> &str {
+        statement
+            .action
+            .first()
+            .and_then(|action| action.split(':').next())
+            .unwrap_or("")
     }
 
     /// Collect all statements from multiple policies
@@ -727,16 +730,6 @@ mod tests {
         )
     }
 
-    /// Extract the service prefix of a statement's first action (e.g. "s3").
-    /// Returns an empty string when the statement has no actions.
-    fn service_prefix(statement: &Statement) -> &str {
-        statement
-            .action
-            .first()
-            .and_then(|action| action.split(':').next())
-            .unwrap_or("")
-    }
-
     #[test]
     fn test_statements_sorted_by_service() {
         let merger = PolicyMerger::new();
@@ -769,7 +762,7 @@ mod tests {
         let services: Vec<&str> = merged_policies[0]
             .statements
             .iter()
-            .map(service_prefix)
+            .map(PolicyMerger::service_prefix)
             .collect();
         assert_eq!(
             services,
@@ -777,10 +770,22 @@ mod tests {
             "statements must be ordered by service prefix with same-service statements adjacent"
         );
 
-        // Sorting must be deterministic across repeated runs over the same input.
-        let rerun = merger.merge_statements(&statements).unwrap();
-        let rerun_services: Vec<&str> = rerun[0].statements.iter().map(service_prefix).collect();
-        assert_eq!(services, rerun_services, "sort order must be stable");
+        // The output must be independent of input order: feeding the same set
+        // in a different (scrambled) order must produce the identical merged
+        // result, statement-for-statement. Comparing the full statements (not
+        // just service prefixes) catches ordering regressions the prefix check
+        // would miss, e.g. the two same-service s3 statements swapping places.
+        let scrambled = vec![
+            statements[2].clone(), // s3:GetObject
+            statements[0].clone(), // s3:PutObject
+            statements[3].clone(), // kms:Decrypt
+            statements[1].clone(), // dynamodb:GetItem
+        ];
+        let scrambled_policies = merger.merge_statements(&scrambled).unwrap();
+        assert_eq!(
+            scrambled_policies[0].statements, merged_policies[0].statements,
+            "sort order must depend only on statement content, not input order"
+        );
     }
 
     #[test]
@@ -826,7 +831,7 @@ mod tests {
         // order. The concatenation must be globally non-decreasing by service.
         let services: Vec<&str> = merged_policies
             .iter()
-            .flat_map(|policy| policy.statements.iter().map(service_prefix))
+            .flat_map(|policy| policy.statements.iter().map(PolicyMerger::service_prefix))
             .collect();
         let mut sorted = services.clone();
         sorted.sort_unstable();
--- a/iam-policy-autopilot-policy-generation/src/policy_generation/merge.rs
+++ b/iam-policy-autopilot-policy-generation/src/policy_generation/merge.rs
@@ -67,7 +67,16 @@ impl ResourceGroup {
 /// Configuration for policy merging behavior
 #[derive(Debug, Clone, Default)]
 pub struct PolicyMergerConfig {
-    /// Allow merging actions from different services into the same statement
+    /// Allow merging actions from different services into the same statement.
+    ///
+    /// This trades readability for compactness. When enabled, statements are
+    /// grouped by shared resource rather than kept one-service-per-statement,
+    /// so a single service's actions may be spread across multiple statements.
+    /// In that mode the by-service statement ordering done by
+    /// `sort_statements_by_service` no longer keeps every service's actions
+    /// adjacent — a cross-service statement is sorted by the service of its
+    /// first action only, so the remaining services' actions can land in a
+    /// non-adjacent statement.
     pub(crate) allow_cross_service_merging: bool,
 }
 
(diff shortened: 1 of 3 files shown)


iam-policy-autopilot-policy-generation/src/policy_generation/merge.rs
```diff
@@ -196,6 +196,11 @@ impl PolicyMerger {
             policies.push(IamPolicy::new());
         }
 
+        // Sort each policy's statements by service for deterministic, review-friendly output
+        for policy in &mut policies {
+            Self::sort_statements_by_service(&mut policy.statements);
+        }
```
mschlaipfer: While deterministic, note that this sorts each policy individually. If two statements from a service were assigned to different policies, this would keep them apart, which could lead to oversight in a human review?

Claude told me that we could sort before assigning statements to policies. This could still lead to services getting split across policies, but it might be clearer. We could in the future add a flag to try to keep a service's statement within one policy (if possible).
```rust
pub(crate) fn merge_statements(&self, statements: &[Statement]) -> Result<Vec<IamPolicy>> {
    if statements.is_empty() {
        return Ok(vec![IamPolicy::new()]);
    }

    // Group statements by mergeable resources with size awareness
    let groups = self.group_statements_by_mergeable_resources(statements)?;

    // Create all merged statements first, then sort globally by service
    let mut merged_statements: Vec<Statement> = groups
        .iter()
        .map(|group| self.create_merged_statement_from_group(group))
        .collect::<Result<Vec<_>>>()?;
    Self::sort_statements_by_service(&mut merged_statements);

    let mut policies = Vec::new();
    let mut current_policy = IamPolicy::new();

    // Distribute pre-sorted statements across policies with size checking
    for merged_statement in merged_statements {
        if self.would_exceed_size_limit(&current_policy, &merged_statement)? {
            if !current_policy.statements.is_empty() {
                policies.push(current_policy);
                current_policy = IamPolicy::new();
            }
            current_policy.add_statement(merged_statement);
            // ... single-statement-too-large warning ...
        } else {
            current_policy.add_statement(merged_statement);
        }
    }

    // ... remainder unchanged (push last policy, ensure non-empty) ...
}

```

mschlaipfer (commented):
Thanks for the PR. Looks good, just wondering if we want to sort globally and not per-policy.

iam-policy-autopilot-policy-generation/src/policy_generation/merge.rs
```diff
@@ -674,6 +709,62 @@ mod tests {
...
+        assert_eq!(
+            services,
+            vec!["dynamodb", "kms", "s3", "s3"],
+            "statements must be ordered by service prefix with same-service statements adjacent"
+        );
+
+        // Sorting must be deterministic across repeated runs over the same input.
+        let rerun = merger.merge_statements(&statements).unwrap();
+        let rerun_services: Vec<&str> = rerun[0].statements.iter().map(service_prefix).collect();
+        assert_eq!(services, rerun_services, "sort order must be stable");
```
adpaco: This check isn't particularly strong. We're executing on the same input and don't compare the full output. I suggest we shuffle the input somehow, then assert that the result is equal to `merged_policies[0].statements`.

iam-policy-autopilot-policy-generation/src/policy_generation/merge.rs
```diff
@@ -206,6 +211,36 @@ impl PolicyMerger {
...
+    /// An empty action list is treated as an empty-string service key so the
+    /// function never panics on malformed statements.
+    fn sort_statements_by_service(statements: &mut [Statement]) {
+        fn service(statement: &Statement) -> &str {
+            statement
+                .action
+                .first()
+                .and_then(|action| action.split(':').next())
+                .unwrap_or("")
+        }
```
adpaco: This is duplicated in `service_prefix` - why not use that?

adpaco (commented):
Thank you! This feature will definitely help understand updates to permissions caused by application code changes.

I agree with @mschlaipfer that we should sort statements globally.

valter-silva-au:
Thanks both — agreed. Pushed an update that sorts the full set of merged statements by service **before** distributing them across policies, instead of sorting each policy after the split.

- `merge_statements` now builds all merged statements, calls `sort_statements_by_service` once on the whole set, then runs the existing size-aware distribution loop over the pre-sorted statements.
- This is essentially the shape @mschlaipfer sketched: sort globally up front. A service can still be split across policies when size limits force it, but the statements stay globally service-ordered and adjacent rather than scattered.
- Added `test_statements_sorted_by_service_across_policy_split`, which interleaves two services with resources large enough to force a multi-policy split and asserts the concatenation across all policies is globally non-decreasing by service. It fails under the old per-policy sort and passes now.

On the future "keep a service within one policy when possible" flag — happy to leave that as a follow-up; the global pre-sort is the natural foundation for it. `cargo fmt`, `clippy --workspace -D warnings`, and `cargo test --workspace` are all green locally. Rebased onto latest `main` to clear the conflict.

iam-policy-autopilot-policy-generation/src/policy_generation/merge.rs
valter-silva-au: Done — implemented this. `merge_statements` now creates all merged statements, sorts the whole set with `sort_statements_by_service`, then runs the size-aware distribution loop over the pre-sorted statements (matching the snippet you shared). Statements no longer get torn across the policy boundary by service. New `test_statements_sorted_by_service_across_policy_split` covers the multi-policy split case.

mschlaipfer:
We require [signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

adpaco:
There is an edge case when `allow_cross_service_merging = true` that is causing this to break the "same service stays together" guarantee written in the changelog. For example, if we have these 4 statements:

```
Statement A: actions = ["s3:GetObject"],       resource = ["*"]
Statement B: actions = ["dynamodb:GetItem"],   resource = ["*"]
Statement C: actions = ["kms:Decrypt"],        resource = ["*"]
Statement D: actions = ["kms:GenerateDataKey"], resource = ["arn:aws:kms:...:key/xyz"]
```

When cross-service is enabled, we will merge actions A, B and C because they share the same resource (picking `dynamodb` as the sort key):

| Position | Sort key | Actions in statement |
|----------|----------|----------------------|
| 1 | `dynamodb` | `dynamodb:GetItem`, `kms:Decrypt`, `s3:GetObject` |
| 2 | `kms` | `kms:GenerateDataKey` |

So the two `kms` actions end up not being together. I understand this is the result of using an off-by-default option but users might expect the guarantee even then. I'd either find a way to resolve this issue or call out the incompatibility in the changelog.

mschlaipfer:
Nice find @adpaco. I think the solution here is to document the behavior on `allow_cross_service_merging`. This option is there to generate more compact policies.

We could also remove the option, but I'd opt against that, as long as there's no correctness issue with the ordering.

valter-silva-au:
Thanks @adpaco @mschlaipfer — addressed both points.

- **Cross-service-merging caveat documented.** Per your call to keep the option and document the behavior, the CHANGELOG entry for the global service sort now spells out that when `--minimize-policy-size` / `allow_cross_service_merging = true` is enabled, statements are grouped by shared resource rather than by service, so the "same service stays together" guarantee doesn't hold for that mode (the example @adpaco gave). The option stays as-is — it exists to produce more compact policies — and the sort remains deterministic.
- **Strengthened the sort test.** The determinism test now feeds the statements in scrambled order and asserts the merged output is identical statement-for-statement to the canonical run, rather than re-running the same input — which catches same-service reordering the prefix-only check missed. Also deduped the `service_prefix` helper into a single associated function used by both the sort and the tests.

Commits are signed. The macOS x86_64 check is failing on an unrelated `enrichment::engine` test (`codeguru-reviewer` service-reference parse — empty file / EOF), not touched by this PR; Linux and Windows are green. Ready for another look.

mschlaipfer (approved):
Let's merge this. I might address some minor things as part of https://github.com/awslabs/iam-policy-autopilot/pull/230/changes

janstey-amzn (approved):

PR #230, opened 2026-06-09
fix: make policy generation output deterministic

mschlaipfer:
## Description

I have started converting tests to use [snapbox](https://github.com/assert-rs/snapbox/). This is in preparation to moving `terraform_integration.rs` over. There may be more changes needed to make everything deterministic.

Replace `HashMap` with `BTreeMap` for `TerraformResources` and `ResolvedResourceMap` so iteration order is stable. Sort FAS-expanded operations and merged-statement conditions. Add `Ord` derives to `Condition`/`Operator`.

## Checklist

- [X] I have updated [`CHANGELOG.md`](https://github.com/awslabs/iam-policy-autopilot/blob/main/CHANGELOG.md) under `[Unreleased]` (if this is a user-facing change)
- [X] Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and use [clear messages](https://github.com/awslabs/iam-policy-autopilot/blob/main/CONTRIBUTING.md#commit-message-guidelines)

---


iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -228,7 +228,10 @@ impl ResourceMatcher {
 
         let mut enriched_actions = vec![];
 
-        for op in fas_expansion.operations() {
+        let mut sorted_ops: Vec<_> = fas_expansion.operations().collect();
+        sorted_ops.sort_by_key(|op| op.service_operation_name());
```
adpaco, line 232: `service_operation_name` allocates a `String` so it would be nicer to use `sort_by_cached_key` here. Not a big deal in this case since the element count will typically be small.

adpaco (approved):
What is the relation/plan with respect to #229 - are the changes in that PR compatible with this?

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
mschlaipfer: Wasn't aware of `sort_by_cached_key`, thanks. Will fix.

mschlaipfer:
#229 sorts statements by action. With that, sorting operations in `resource_matching.rs` may not be required. Other changes affect ordering of the explanations for resource binding from Terraform.

I'm happy to apply this change after #229 is in (and simplify).

This change came from experimenting with the `snapshot` crate, for which I needed deterministic output.

PR #232, opened 2026-06-12
feat: support session-based client/resource creation in variable type tracking

adpaco:
Closes #226

## Description

Extends the variable type tracker to handle `boto3.Session().client('service')` and `boto3.Session().resource('service')` patterns, which are the recommended boto3 usage in production code (especially Lambda handlers and multi-region deployments).

### What's New

- **Session variable detection**: Tracks `session = boto3.Session(...)` assignments at module and function level
- **Session factory matching**: Resolves `session.client('s3')` / `session.resource('dynamodb')` to the correct service
- **Scope-aware lookup**: Session variables are scoped per-function -> a `boto3.Session()` in one function doesn't validate `.client()` calls in other functions
- **Local shadowing detection**: If a function locally reassigns a session variable name (e.g., `session = requests.Session()`), the module-level `boto3.Session()` fallback is skipped

### Example

```python
import boto3

session = boto3.Session(region_name='us-east-1')
acm_client = session.client('acm')
# list_certificates exists in acm, iot, AND transfer (all with no required params)
response = acm_client.list_certificates()
```

**Before (no session tracking):** `PossibleServices: ["acm", "iot", "transfer"]` -> 3 policy statements
**After (with session tracking):** `PossibleServices: ["acm"]` -> 1 policy statement

### Testing

- 61 unit tests (parameterized with rstest)
- Covers: session client/resource, extra kwargs, function-scoped sessions, non-session rejection, cross-scope isolation, local shadowing

## Checklist

- [x] I have updated [`CHANGELOG.md`](https://github.com/awslabs/iam-policy-autopilot/blob/main/CHANGELOG.md) under `[Unreleased]` (if this is a user-facing change)
- [x] Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and use [clear messages](https://github.com/awslabs/iam-policy-autopilot/blob/main/CONTRIBUTING.md#commit-message-guidelines)

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #199, opened 2026-05-12
fix: merge duplicate condition values in `serialize_conditions`

adpaco:
*Description of changes:*

Fixed `serialize_conditions` silently dropping condition values when two `Condition` structs share the same operator and key. The `HashMap::insert` overwrote the first entry instead of merging values, producing more restrictive IAM policies than intended. Added parameterized `rstest` tests covering duplicate operator+key (the bug), different keys under same operator, mixed operators on same key, and empty condition values.

Without the changes included in this PR, case 1 and 4 fail:

```
running 4 tests
test policy_generation::tests::test_serialize_conditions::case_3_mixed_operators_same_key ... ok
test policy_generation::tests::test_serialize_conditions::case_2_different_keys_same_operator ... ok
test policy_generation::tests::test_serialize_conditions::case_1_duplicate_operator_key_merges_values ... FAILED
test policy_generation::tests::test_serialize_conditions::case_4_empty_values ... FAILED

failures:

---- policy_generation::tests::test_serialize_conditions::case_1_duplicate_operator_key_merges_values stdout ----

thread 'policy_generation::tests::test_serialize_conditions::case_1_duplicate_operator_key_merges_values' (6796909) panicked at iam-policy-autopilot-policy-generation/src/policy_generation/mod.rs:383:9:
assertion `left == right` failed: Condition block mismatch.
  actual: {"StringEquals":{"kms:ViaService":["dynamodb.us-east-1.amazonaws.com"]}}
  expected: {"StringEquals":{"kms:ViaService":["s3.us-east-1.amazonaws.com","dynamodb.us-east-1.amazonaws.com"]}}
  left: Object {"StringEquals": Object {"kms:ViaService": Array [String("dynamodb.us-east-1.amazonaws.com")]}}
 right: Object {"StringEquals": Object {"kms:ViaService": Array [String("dynamodb.us-east-1.amazonaws.com"), String("s3.us-east-1.amazonaws.com")]}}
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- policy_generation::tests::test_serialize_conditions::case_4_empty_values stdout ----

thread 'policy_generation::tests::test_serialize_conditions::case_4_empty_values' (6796912) panicked at iam-policy-autopilot-policy-generation/src/policy_generation/mod.rs:383:9:
assertion `left == right` failed: Condition block mismatch.
  actual: {"StringEquals":{"kms:ViaService":[]}}
[...]

PR #235, opened 2026-06-17
chore: bump version to 0.2.3

ChrisHsu12:
### Added

- Variable type tracking for boto3 clients and resources — improves extraction precision when clients are passed across function boundaries (#128)
- Support for `boto3.Session().client()` / `.resource()` patterns in variable type tracking (#232)
- `--resource-cutoff` CLI flag and `resource_cutoff` MCP input to configure when resource lists collapse to `*` (#217)
- Support for namespace imports in TypeScript/JavaScript (#190)
- Added partial support for permissions needed by [aws-lambda-powertools](https://pypi.org/project/aws-lambda-powertools/) (#186)
### Fixed
- We now respect the system's native certificate store instead of using bundled certificates (#209)
- `--explain` now shows every call site when the same operation appears multiple times (#188)
- Condition values for the same key are now merged instead of overwritten when serializing policies (#199)
### Changed
- `EnrichmentEngine::new` now requires a `resource_cutoff` parameter; use `DEFAULT_RESOURCE_CUTOFF` to preserve existing behavior (#217)

## Checklist

- [X] I have updated [`CHANGELOG.md`](https://github.com/awslabs/iam-policy-autopilot/blob/main/CHANGELOG.md) under `[Unreleased]` (if this is a user-facing change)
- [ ] Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and use [clear messages](https://github.com/awslabs/iam-policy-autopilot/blob/main/CONTRIBUTING.md#commit-message-guidelines)

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-## [Unreleased]
+## Unreleased
+
+## [0.2.3] - 2026-06-17
 
 ### Added
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ exclude = ["iam-policy-autopilot-lints"]
 resolver = "2"
 
 [workspace.package]
-version = "0.2.2"
+version = "0.2.3"
 edition = "2021"
 license = "Apache-2.0"
 repository = "https://github.com/awslabs/iam-policy-autopilot"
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
-## [Unreleased]
+## Unreleased
```
janstey-amzn, line 1: nit: just missing the brackets

janstey-amzn (approved):

janstey-amzn:
You need signed commits: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #237, opened 2026-06-18
feat: support resource-operation chains

mschlaipfer:
Closes #

## Description

Added support for boto3 resources with chained operation calls like `s3.Object("my-key").put(Body="data")` and sub-resources like `s3.Bucket("my-bucket").Object("my-key").put(Body="data")`.

Removed the fall-back that would extract all operations of a resource if the operation couldn't be found (as in `s3.Object("my-key")` without a `.put(..)`).

## Checklist

---

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/extraction/python/resource_direct_calls_extractor.rs
+++ b/iam-policy-autopilot-policy-generation/src/extraction/python/resource_direct_calls_extractor.rs
@@ -38,6 +38,7 @@ use crate::extraction::python::boto3_resources_model::{
     Boto3ResourcesModel, Boto3ResourcesRegistry, HasManySpec, OperationType,
 };
 use crate::extraction::python::common::ArgumentExtractor;
+use crate::extraction::python::node_kinds;
 use crate::extraction::{
     AstWithSourceFile, Parameter, ParameterValue, SdkMethodCall, SdkMethodCallMetadata,
 };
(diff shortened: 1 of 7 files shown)


iam-policy-autopilot-policy-generation/tests/python_chained_subresource_action_test.rs
```diff
@@ -0,0 +1,143 @@
...
+#[case::assignment_only_no_action(
+    r#"
+import boto3
+
+def handle():
+    s3 = boto3.resource("s3")
+    obj = s3.Bucket("my-bucket").Object("my-key")
+"#,
+    &[]
+)]
```
janstey-amzn, line 131: Is it worth adding more test cases here  
 
- A deeper chain (3+ levels), `s3.Object("b", "k").MultipartUpload("id").Part(1).upload(...)`  verifies the recursive accumulation actually works beyond depth 2
- A chain where the accessor name differs from the target type (`bucket.Acl()` -> type **BucketAcl**) to verify the sub_resources keying by accessor works
- An invalid chain (`s3.Bucket("b").NonexistentResource("x").put(...)`) to verify it produces nothing

janstey-amzn (dismissed):

iam-policy-autopilot-policy-generation/src/extraction/python/resource_direct_calls_extractor.rs
```diff
@@ -574,6 +434,315 @@ impl<'a> ResourceDirectCallsExtractor<'a> {
...
+        let mut method_calls = Vec::new();
+
+        // First pass: resolve assignments whose RHS is a nested sub-resource chain, e.g.
+        // `obj = s3.Bucket("b").Object("k")`. Maps the assigned variable name to the
+        // resolved leaf resource so later `obj.method(...)` calls can be matched.
+        let assigned_resources = self.resolve_assigned_resource_chains(ast, registry);
+
+        // Second pass: every call `<receiver>.method(args)`. The receiver is either an
+        // inline nested chain or a variable bound to one (from the first pass).
+        for node_match in root.find_all("$RECEIVER.$METHOD($$$ARGS)") {
```
adpaco, line 473: Wouldn't this match smaller, intermediate calls in the chain? In that case, we could save some work by checking if the method name is a known sub-resource.

adpaco (approved):

nwinans-aws (approved):

janstey-amzn (approved):

PR #257, opened 2026-07-24
feat: Enumerated-action regression tests, wildcard-resource warnings, and shape telemetry

janstey-amzn:
### Description

Hardens the enumerated-actions property of policy generation and surfaces wildcard-resource fallbacks to callers.

**Regression tests for wildcard-free actions.** Generated policies never contain a wildcard in any `Action` entry, neither a bare `"*"` nor an embedded wildcard like `"s3:*"`. New integration tests pin this across multi-service generation, policy merging (with and without `--minimize-policy-size`), and the `Resource: "*"` fallback paths. An end-to-end test additionally mocks the Service Reference endpoint with a small exhaustive action catalog, exercises every catalog action through the real enrichment pipeline, and asserts the merged output still enumerates each action individually instead of collapsing to a service wildcard. A mutation check (temporarily patching the merge logic to collapse same-service actions) confirmed the tests detect the failure mode.

**`Warnings` field in the generation result.** Statements whose `Resource` falls back to the `"*"` wildcard (no ARN patterns available, an empty resource list, or the resource list collapsed by the resource cutoff) are now flagged in a `Warnings` array on `GeneratePoliciesResult`. Each warning carries a machine-recognizable `WarningType` enum (`#[non_exhaustive]`, so future categories can be added without breaking consumers) plus the statement's policy index, SID, and actions, letting callers construct their own review messages. A plain-English `message` is included for CLI users. Warnings are computed against the final (post-merge) policies so indices match the returned output.

**Telemetry shape metrics.** Three new result fields establish an anomaly baseline for generated policy size: `num_statements_generated`, `num_actions_generated`, and `num_wildcard_resource_statements`. Counts only, no policy content is collected. `TELEMETRY.md` is updated (enforced by the doc-sync tests).

**Testing:** full workspace suite passes (1256 tests), `cargo clippy --workspace -- -D warnings` clean, `cargo fmt --check` clean.

## Checklist

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,6 @@
 
 - Support for chained and nested boto3 sub-resource actions, including calls on a variable bound to a chain — e.g. `s3.Bucket("b").put_object(...)`, `s3.Bucket("b").Object("k").put(...)`, and `obj = s3.Bucket("b").Object("k"); obj.put(...)` now resolve to the underlying operation with identifiers injected from the chain
 - `Warnings` field in the `generate-policies` output flagging statements whose `Resource` fell back to the `"*"` wildcard (no ARN patterns available, an empty resource list, or the resource list was collapsed by the resource cutoff). Each warning carries a machine-recognizable `WarningType` plus the statement's policy index, SID, and actions, so callers can construct their own review messages
-- New telemetry result metrics for generation runs: `num_statements_generated`, `num_actions_generated`, and `num_wildcard_resource_statements` (counts only, no policy content). See [TELEMETRY.md](TELEMETRY.md)
-- Regression tests pinning that generated policies never contain a wildcard in any `Action` entry — actions are always fully enumerated, even when the analyzed code uses every action a service defines
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,11 @@
 ## [Unreleased]
 
 ### Added
+
 - Support for chained and nested boto3 sub-resource actions, including calls on a variable bound to a chain — e.g. `s3.Bucket("b").put_object(...)`, `s3.Bucket("b").Object("k").put(...)`, and `obj = s3.Bucket("b").Object("k"); obj.put(...)` now resolve to the underlying operation with identifiers injected from the chain
+- `Warnings` field in the `generate-policies` output flagging statements whose `Resource` fell back to the `"*"` wildcard (no ARN patterns available, an empty resource list, or the resource list was collapsed by the resource cutoff). Each warning carries a machine-recognizable `WarningType` plus the statement's policy index, SID, and actions, so callers can construct their own review messages
+- New telemetry result metrics for generation runs: `num_statements_generated`, `num_actions_generated`, and `num_wildcard_resource_statements` (counts only, no policy content). See [TELEMETRY.md](TELEMETRY.md)
+- Regression tests pinning that generated policies never contain a wildcard in any `Action` entry — actions are always fully enumerated, even when the analyzed code uses every action a service defines
 
 ### Changed
 
@@ -56,7 +60,7 @@
 ### Added
 
 - IAM Policy Autopilot now supports policy generation for Java applications. (#134)
-- When provided with Terraform configurations or plans, IAM Policy Autopilot now generates more precise resource blocks, e.g., narrowing arn:aws:s3:::* down to the actual bucket/resource referenced. (#157)
+- When provided with Terraform configurations or plans, IAM Policy Autopilot now generates more precise resource blocks, e.g., narrowing arn:aws:s3:::\* down to the actual bucket/resource referenced. (#157)
 - IAM Policy Autopilot now supports overriding the default HTTP bind address of the MCP server. (#159)
 - This release adds anonymous usage telemetry. Set IAM_POLICY_AUTOPILOT_TELEMETRY=0 to disable. See TELEMETRY.md for details (#174)
 
@@ -96,7 +100,7 @@
 
 ### Changed
 
-- We now add the policy ID `IamPolicyAutopilot` in the access denied workflow.  (#48)
+- We now add the policy ID `IamPolicyAutopilot` in the access denied workflow. (#48)
 - Updated Cargo.toml description. (#46)
 
 ## [0.1.1] - 2025-11-26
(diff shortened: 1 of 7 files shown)


iam-policy-autopilot-policy-generation/src/policy_generation/integration_tests.rs
```diff
@@ -18,6 +21,454 @@ mod tests {
...
+            }
+        }
+        assert!(
+            actions_checked > 0,
+            "no actions were checked — the test input produced no statements"
+        );
+    }
+
+    /// Convenience wrapper for a full generation result.
+    fn assert_result_has_no_wildcard_actions(result: &GeneratePoliciesResult) {
```
mschlaipfer: doesn't provide much convenience, I think this should be removed or have `assert_no_wildcard_actions` take `&GeneratePoliciesResult`

iam-policy-autopilot-policy-generation/src/policy_generation/integration_tests.rs
```diff
@@ -18,6 +21,454 @@ mod tests {
...
+                 no more and no fewer \
+                 (allow_cross_service_merging={allow_cross_service_merging})"
+            );
+        }
+    }
+
+    /// Regression test: a generation run over many SDK calls across multiple
+    /// services never emits a wildcard Action.
+    #[test]
+    fn test_actions_are_never_wildcards_multi_service_generation() {
```
mschlaipfer: Can we define a property/properties we want to test and then use rstest parametrized tests for the different inputs we want to test?

CHANGELOG.md
```diff
@@ -1,7 +1,11 @@
 ## [Unreleased]
 
 ### Added
+
 - Support for chained and nested boto3 sub-resource actions, including calls on a variable bound to a chain — e.g. `s3.Bucket("b").put_object(...)`, `s3.Bucket("b").Object("k").put(...)`, and `obj = s3.Bucket("b").Object("k"); obj.put(...)` now resolve to the underlying operation with identifiers injected from the chain
+- `Warnings` field in the `generate-policies` output flagging statements whose `Resource` fell back to the `"*"` wildcard (no ARN patterns available, an empty resource list, or the resource list was collapsed by the resource cutoff). Each warning carries a machine-recognizable `WarningType` plus the statement's policy index, SID, and actions, so callers can construct their own review messages
+- New telemetry result metrics for generation runs: `num_statements_generated`, `num_actions_generated`, and `num_wildcard_resource_statements` (counts only, no policy content). See [TELEMETRY.md](TELEMETRY.md)
+- Regression tests pinning that generated policies never contain a wildcard in any `Action` entry — actions are always fully enumerated, even when the analyzed code uses every action a service defines
```
mschlaipfer: Not user-facing, so we should remove it from the CHANGELOG

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -61,6 +61,73 @@ pub struct GeneratePoliciesResult {
...
+    /// Machine-recognizable warning category
+    pub warning_type: PolicyWarningType,
+    /// Index of the policy in `policies` that contains the statement
+    pub policy_index: usize,
+    /// SID of the statement, when present. Merged statements have no SID;
+    /// use `policy_index` and `actions` to locate them.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub sid: Option<String>,
+    /// Actions of the flagged statement
+    pub actions: Vec<String>,
```
mschlaipfer: How are the actions used?

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -61,6 +61,73 @@ pub struct GeneratePoliciesResult {
...
+#[serde(rename_all = "PascalCase")]
+pub struct PolicyWarning {
+    /// Machine-recognizable warning category
+    pub warning_type: PolicyWarningType,
+    /// Index of the policy in `policies` that contains the statement
+    pub policy_index: usize,
+    /// SID of the statement, when present. Merged statements have no SID;
+    /// use `policy_index` and `actions` to locate them.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub sid: Option<String>,
```
mschlaipfer: I think `sid` doesn't help here.

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -61,6 +61,73 @@ pub struct GeneratePoliciesResult {
...
+/// convenience English rendering for CLI users. Callers producing their own
+/// messages should use `warning_type` plus the statement metadata
+/// (`policy_index`, `sid`, `actions`) instead of `message`.
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct PolicyWarning {
+    /// Machine-recognizable warning category
+    pub warning_type: PolicyWarningType,
+    /// Index of the policy in `policies` that contains the statement
+    pub policy_index: usize,
```
mschlaipfer: Would it be better to use a path expression (I think there are different ways to express JSON paths, not sure if there are tradeoffs/which is best) which allows to navigate to the statement in the generated policy?

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -146,6 +213,93 @@ impl AwsContext {
...
+            policy.add_statement(statement);
+        }
+        PolicyWithMetadata {
+            policy,
+            policy_type: PolicyType::Identity,
+        }
+    }
+
+    #[test]
+    fn test_wildcard_resource_warnings_flags_wildcard_statements() {
```
mschlaipfer: Can we make all these unit tests into one parametrized test using rstest where the expected value is a `Vec<PolicyWarning>`?

iam-policy-autopilot-policy-generation/src/policy_generation/engine.rs
```diff
@@ -275,10 +275,15 @@ impl<'a> Engine<'a> {
         // Collect explanations
         let explanations = extract_explanations(enriched_calls);
 
+        // Flag statements that fell back to Resource "*" so consumers can
+        // surface them for review
+        let warnings = crate::api::model::PolicyWarning::wildcard_resource_warnings(&policies);
```
mschlaipfer: NP: Can we import `crate::api::model::PolicyWarning::wildcard_resource_warnings`

iam-policy-autopilot-policy-generation/src/api/model.rs
janstey-amzn: JSON Path is good for programatic mapping, I think an index is more universal for humans to understand as well

mschlaipfer: Are we sure that we don't want more precise navigation? Is this format forwards-compatible if we want to make it more precise?

CHANGELOG.md
```diff
@@ -56,7 +58,7 @@
 ### Added
 
 - IAM Policy Autopilot now supports policy generation for Java applications. (#134)
-- When provided with Terraform configurations or plans, IAM Policy Autopilot now generates more precise resource blocks, e.g., narrowing arn:aws:s3:::* down to the actual bucket/resource referenced. (#157)
+- When provided with Terraform configurations or plans, IAM Policy Autopilot now generates more precise resource blocks, e.g., narrowing arn:aws:s3:::\* down to the actual bucket/resource referenced. (#157)
```
mschlaipfer, line 62: I think we should add new telemetry being collected to the CHANGELOG to be transparent.

janstey-amzn: Done

mschlaipfer (approved):

PR #261, opened 2026-07-31
feat: include service reference modified dates in policy output for auditability

### LINKED ISSUE #260: Include service reference modfied dates into IPAs output for auditability
Service reference publishes a modified timestamp per service (https://servicereference.us-east-1.amazonaws.com/). For auditability we should include this in IPA's output. Note that we need to add this data for each service contributing to the generated policy.

Today, the service reference endpoint doesn't allow querying previous versions, however.

arimu1:
Closes #260

## Description

Service reference publishes a per-service `modified` epoch timestamp at https://servicereference.us-east-1.amazonaws.com/. This change includes those timestamps in IPA's `generate-policies` output for each service that contributes actions to the generated policy, so results can be audited against the service-reference versions that informed them.

**Changes:**
- Parse `modified` from the service reference index (defaults to `0` when absent for backward compatibility)
- Add `ServiceReferenceModified` (`BTreeMap<service, epoch_seconds>`) to `GeneratePoliciesResult`
- Populate it from action prefixes in the final policies after generation/merge
- Best-effort: index fetch failures log a warning and omit the field rather than failing generation

**Example output shape:**
```json
{
  "Policies": [ ... ],
  "ServiceReferenceModified": {
    "ec2": 1785355296,
    "s3": 1774454984
  }
}
```

## Testing
- `cargo test -p iam-policy-autopilot-policy-generation --lib enrichment::service_reference` (22 passed, 2 ignored network)
- `cargo test -p iam-policy-autopilot-policy-generation --lib -- services_from_policies service_reference_modified modified_for_services`
- `cargo test -p iam-policy-autopilot-policy-generation --lib enrichment::` (144 passed)
- `cargo test -p iam-policy-autopilot-policy-generation --lib policy_generation::` (102 passed)
- `cargo test -p iam-policy-autopilot-mcp-server --lib` (22 passed)
- `cargo fmt --all -- --check`
- `cargo clippy -p iam-policy-autopilot-policy-generation --lib -- -D warnings`

## Checklist

- [ ] Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and use [clear messages](https://github.com/awslabs/iam-policy-autopilot/blob/main/CONTRIBUTING.md#commit-message-guidelines)

---

### DIFF
--- a/iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
+++ b/iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
@@ -227,6 +227,7 @@ pub async fn generate_policies(config: &GeneratePolicyConfig) -> Result<Generate
             explanations: None,
             resource_binding_explanations: None,
             warnings: vec![],
+            service_reference_modified: BTreeMap::new(),
         });
     }
 
@@ -269,6 +270,7 @@ pub async fn generate_policies(config: &GeneratePolicyConfig) -> Result<Generate
             explanations: None,
             resource_binding_explanations: None,
             warnings: vec![],
+            service_reference_modified: BTreeMap::new(),
         });
     }
 
@@ -350,6 +352,12 @@ pub async fn generate_policies(config: &GeneratePolicyConfig) -> Result<Generate
     // policy indices and statements match what the caller receives
     let warnings = crate::api::model::PolicyWarning::wildcard_resource_warnings(&final_policies);
 
+    // Service-reference modified timestamps were captured at enrichment time
+    // (on each EnrichedSdkMethodCall) and aggregated in the policy engine —
+    // same pattern as explanations. Merging policies does not change which
+    // services informed enrichment, so no recompute is needed.
+    let service_reference_modified = result.service_reference_modified;
+
     iam_policy_autopilot_common::telemetry::span::record_result_number(
         "num_policies_generated",
         final_policies.len(),
@@ -383,6 +391,7 @@ pub async fn generate_policies(config: &GeneratePolicyConfig) -> Result<Generate
         explanations,
         resource_binding_explanations: binding_explanations,
         warnings,
+        service_reference_modified,
     })
 }
 
--- a/iam-policy-autopilot-policy-generation/src/api/model.rs
+++ b/iam-policy-autopilot-policy-generation/src/api/model.rs
@@ -6,6 +6,7 @@ use crate::{
     enrichment::Explanations, policy_generation::PolicyWithMetadata,
 };
 use anyhow::{anyhow, Result};
+use std::collections::BTreeMap;
 use std::path::PathBuf;
 
 /// Configuration for generate_policies API
@@ -64,6 +65,12 @@ pub struct GeneratePoliciesResult {
     /// Warnings about statements that could not be scoped to specific resources
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub warnings: Vec<PolicyWarning>,
+    /// Epoch-second `modified` timestamps from the service reference index
+    /// (https://servicereference.us-east-1.amazonaws.com/) for each service that
+    /// contributes actions to the generated policies. Enables auditability of
+    /// which service-reference versions informed the result.
+    #[serde(skip_serializing_if = "BTreeMap::is_empty")]
+    pub service_reference_modified: BTreeMap<String, u64>,
 }
 
 /// Machine-recognizable category of a [`PolicyWarning`].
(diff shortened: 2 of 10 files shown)


janstey-amzn:
> Commits must have verified signatures. 

Please see our [CONTRIBUTING.md](https://github.com/awslabs/iam-policy-autopilot/blob/main/CONTRIBUTING.md) for how to set this up, and follow this package's best practices.

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
```diff
@@ -350,6 +352,13 @@ pub async fn generate_policies(config: &GeneratePolicyConfig) -> Result<Generate
     // policy indices and statements match what the caller receives
     let warnings = crate::api::model::PolicyWarning::wildcard_resource_warnings(&final_policies);
 
+    // Per-service service-reference modified timestamps for auditability
+    let contributing_services = crate::api::model::services_from_policies(&final_policies);
```
mschlaipfer: We should not compute this data from the generated policies, but save this information when we enrich an operation. The information should be added to https://github.com/awslabs/iam-policy-autopilot/blob/dfca9373d0e867120d07ba75ebc81e9a27dd6625/iam-policy-autopilot-policy-generation/src/enrichment/mod.rs#L273 and we should generate the right data structure similar to how we generate the explanations: https://github.com/awslabs/iam-policy-autopilot/blob/dfca9373d0e867120d07ba75ebc81e9a27dd6625/iam-policy-autopilot-policy-generation/src/policy_generation/engine.rs#L291

iam-policy-autopilot-policy-generation/src/api/model.rs
```diff
@@ -64,6 +65,31 @@ pub struct GeneratePoliciesResult {
...
+                if let Some((service, _)) = action.split_once(':') {
+                    if !service.is_empty() {
+                        services.insert(service.to_string());
+                    }
+                }
+            }
+        }
+    }
+    services.into_iter().collect()
 }
```
janstey-amzn: Nice addition, and deriving this from the final policies is the right call. One suggestion: the nested loops here can be flattened into an iterator chain, and since `modified_for_services` dedupes anyway, we can skip the owned `String`s and just return borrowed slices:
```
pub(crate) fn services_from_policies(policies: &[PolicyWithMetadata]) -> BTreeSet<&str> {
    policies
        .iter()
        .flat_map(|p| &p.policy.statements)
        .flat_map(|s| &s.action)
        .filter_map(|a| a.split_once(':'))
        .map(|(service, _)| service)
        .filter(|s| !s.is_empty())
        .collect()
}
```

Same behavior, just flatter and allocation-free. The test only needs its expected value switched to a BTreeSet.

iam-policy-autopilot-policy-generation/src/api/generate_policies.rs
janstey-amzn: This is more what I was thinking as well, but can we expect any difference after merging policies? Probably not

arimu1:
Thanks for the reviews — addressed as follows:

### @mschlaipfer / @janstey-amzn (architecture)
Service-reference `modified` timestamps are now captured **during enrichment** on each `EnrichedSdkMethodCall` (when the service reference for that service is used), then aggregated in the policy engine via `extract_service_reference_modified` — same pattern as explanations. The post-hoc scan of final policies + `modified_for_services` call in the API layer is gone. Merge does not recompute timestamps (as @janstey-amzn noted, services shouldn’t change).

### @janstey-amzn (`services_from_policies`)
That helper was removed with the architecture change; timestamps no longer depend on scanning final policy actions.

### Signed commits
Tip is now a single **SSH-signed** commit. If GitHub still shows the signature as unverified, the signing public key still needs to be registered on the arimu1 account as an SSH **signing** key (`Settings → SSH and GPG keys → New SSH key → Key type: Signing Key`). Happy to re-push once that’s in place if verification still fails.

iam-policy-autopilot-policy-generation/src/enrichment/service_reference.rs
```diff
@@ -264,21 +264,33 @@ where
...
 fn deserialize_service_reference_mapping(
     value: Value,
-) -> crate::errors::Result<HashMap<String, Url>> {
+) -> crate::errors::Result<HashMap<String, ServiceReferenceEntry>> {
     #[derive(Deserialize)]
     struct ServiceEntry {
         service: String,
         url: String,
+        /// Defaults to 0 for index entries that predate the `modified` field.
+        #[serde(default)]
```
mschlaipfer, line 292: I think we should not have a `serde(default)` here. Let's assume this is present.

I realize this might lead to changes in tests. I don't know what the extent of it is, though.

iam-policy-autopilot-policy-generation/src/enrichment/resource_matcher.rs
```diff
@@ -315,11 +315,19 @@ impl ResourceMatcher {
             return Ok(None);
         }
 
+        // Capture index `modified` while we still know which service reference
+        // document informed this enrichment (audit trail for #260).
+        let service_reference_modified = service_reference_loader
+            .modified_for_service(original_service_name)
+            .await
+            .filter(|&m| m != 0);
```
mschlaipfer, line 323: This won't be `0` anymore once we remove the `serde(default)`, so we can remove this.

iam-policy-autopilot-policy-generation/src/enrichment/mod.rs
```diff
@@ -279,6 +279,11 @@ pub struct EnrichedSdkMethodCall<'a> {
     pub(crate) actions: Vec<Action>,
     /// The initial SDK method call
     pub(crate) sdk_method_call: &'a SdkMethodCall,
+    /// Epoch-second `modified` timestamp from the service reference index for
+    /// [`Self::service`], when available. Captured at enrichment time so policy
+    /// generation can report which service-reference version informed each call.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) service_reference_modified: Option<u64>,
```
mschlaipfer, line 286: This won't need to be an `Option<...>` once we remove the `serde(default)` on the service reference entry struct.

mschlaipfer (commented):
Thanks for the changes! This is on the right way.

I'm realizing this is going beyond "good first issue". Feel free to let us know, and we can take this over.

